PR #36, opened 2014-02-03
Make FakeStrictRedis inherit redis.StrictRedis

clee704:
Some libraries check whether the type of an argument is
redis.StrictRedis or not. If FakeStrictRedis is complete, this
inheritance should make no difference.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -79,7 +79,7 @@ class _StrKeyDict(MutableMapping):
         return copy.deepcopy(self._dict)
 
 
-class FakeStrictRedis(object):
+class FakeStrictRedis(redis.StrictRedis):
     @classmethod
     def from_url(cls, url, db=None, **kwargs):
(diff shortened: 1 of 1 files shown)


btashton:
This would also allow for directly using some of the convenience functions in StrictRedis.  Right now in #73 I have to copy over the iterator code because I cannot do something like:
```
def scan_iter(self, match=None, count=None):
    return redis.StrictRedis.scan_iter(self, match, count)
```

btashton:
On second though this seems like it might not be a great idea because you then have to add NotImplementedError on all of the not supported library functions.

jamesls:
After thinking about it more, I tend to agree with @btashton.  Subclassing would mean we may inadvertently call into the real redis-py module, which is not ideal.  Given it's easy to make your own custom subclass that inherits from both classes, I'd prefer not to have this in fakeredis.

Closing for now.

PR #70, opened 2015-02-19
Add hincbryfloat operation.

enginous:
Adds support for the `hincrbyfloat` operation, raising `ResponseError`s on the same error conditions as redis does through py-redis.


jamesls:
Thanks!

PR #71, opened 2015-02-25
Fix return value from expire and expireat

dairiki:
Expire and expireat should return a boolean indicating whether the timeout was successfully set.  In general, this means they return `True` if the specified key exists, `False` otherwise.


jamesls:
Thanks!

PR #68, opened 2015-01-14
support scan command, return 0 and all keys match the pattern

liamchzh:
http://redis.io/commands/scan

PR #73, opened 2015-03-04
support scan and scan_iter commands

btashton:
This is a continuation of https://github.com/jamesls/fakeredis/pull/68

Adds support and test case for scan_iter.  Also fixes a bug in scan where you could no use a string for the cursor, which is supported by the real redis-py library.  The implementation of scan_iter is exactly what is in redis-py.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -315,14 +315,14 @@ class FakeStrictRedis(object):
     def scan(self, cursor=0, match=None, count=None):
         if count is None:
             count = 10
-        match_keys = [key for key in self._db
-                if not key or not match or
-                fnmatch.fnmatch(to_native(key), to_native(match))]
+        match_keys = [key for key in self._db \
+            if not key or not match or \
+            fnmatch.fnmatch(to_native(key), to_native(match))]
         keys = match_keys[cursor:cursor+count]
         if cursor + count >= len(match_keys):
             return 0, keys
         else:
-            return cursor + count, keys
+            return cursor + count, keys[cursor:cursor+count]
 
     def set(self, name, value, ex=None, px=None, nx=False, xx=False):
         if (not nx and not xx) \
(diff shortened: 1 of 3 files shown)


liamchzh:
LGTM

fakeredis.py
```diff
@@ -312,6 +312,27 @@ def renamenx(self, src, dst):
...
+        if count is None:
+            count = 10
+        match_keys = [key for key in self._db \
+            if not key or not match or \
+            fnmatch.fnmatch(to_native(key), to_native(match))]
+        keys = match_keys[cursor:cursor+count]
+        if cursor + count >= len(match_keys):
+            return 0, keys
+        else:
+            return cursor + count, keys[cursor:cursor+count]
```
jamesls: I believe this should just be `return cursor + count, keys` because you've already sliced the list on line 322.

fakeredis.py
```diff
@@ -312,6 +312,27 @@ def renamenx(self, src, dst):
         else:
             return self.rename(src, dst)
 
+    def scan(self, cursor=0, match=None, count=None):
+        cursor = int(cursor)
+        if count is None:
+            count = 10
+        match_keys = [key for key in self._db \
+            if not key or not match or \
```
jamesls: `if not key` would exclude the empty string which I believe is a valid key name.

jamesls:
Merged in https://github.com/jamesls/fakeredis/commit/f6f6a2f08647405a6ac756952d4ac1523c8a5830 after rebasing against master.

I went ahead and fixed the two issues above (https://github.com/jamesls/fakeredis/commit/bd4d62216fa685c3ad6d33409fe62983d77f9a11) and added a few additional test cases.  Let me know if I missed anything or you have any feedback.

Thanks for the pull request!

PR #80, opened 2015-05-13
add method `incrby` for fakeredis

lfyzjck:
redis-py client support command `incrby` which fakeredis not support.

https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L914


jamesls:
Thanks!

PR #86, opened 2015-07-10
[Fix bug & fix code style] TypeError: zadd() keywords must be strings when using FakeRedis.

fatelei:
When call method `zadd(name, value, score)` and value is integer, it will lead to raise TypeError zadd() keywords must be strings.

For example:

```
from fakeredis import FakeRedis

f = FakeRedis()
f.zadd('foo', 1, 1)
```

The version 0.6.2 will raise TypeError: zadd() keywords must be strings

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -1287,6 +1287,7 @@ class FakeRedis(FakeStrictRedis):
 
         The ``value`` and ``score`` arguments are deprecated.
         """
+        print pairs
         if value is not None or score is not None:
             if value is None or score is None:
                 raise redis.RedisError(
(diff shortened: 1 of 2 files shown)


fatelei:
@jamesls Could you take some time review this pr? Thanks.

jamesls:
Sorry for not reaching out earlier, thanks for the PR.  The only thing blocking this being merged (as per our [contributing guidelines](https://github.com/jamesls/fakeredis/blob/master/CONTRIBUTING.rst)) is a unit test showing what the problem is/verifying the fix.  If you could write a test this can get merged quicker.  Otherwise it will have to wait until I have some time to spare to write a test.  Thanks!

fatelei:
Ok, I will add a test.

fatelei:
@jamesls , I have added related test for this bugfix.

jamesls:
Thanks!

fatelei:
：)

PR #92, opened 2015-09-24
Simulate HyperLogLog functionality by using SETs

nelz9999:
This PR creates API-compatible methods for ```pfadd```, ```pfcount```, and ```pfmerge```

This functionality should not be as performant as actual HyperLogLog implementations. Nor should it be as kind to memory, if using it very hard.

The only "bug" this might not help a user catch is if they actually try to use a SET in a HyperLogLog context.


nelz9999:
Dammit, I forgot to turn off ```yapf``` before sending this in. Gimme a sec.

fakeredis.py
```diff
@@ -1273,6 +1273,42 @@ def transaction(self, func, *keys):
...
+        """
+        Return the approximated cardinality of
+        the set observed by the HyperLogLog at key(s).
+        """
+        return self.scard(name)
+
+        # This will be the proper implementation, once `fakeredis` uses a
+        # post-2.10.3 release of `redis-py`.
+        # https://github.com/andymccurdy/redis-py/commit/6935a30c680a81659b2e02b9f9220517cdfd637b
+        # def pfcount(self, *sources):
```
jamesls: I'd prefer to just get fakeredis updated with the latest version of redis-py.  This should only require a change in the requirements.txt file right?  Would you mind doing this?

nelz9999: I would do it if I could, but the ```redis-py``` package hasn't released anything after 2.10.3 yet, so I have no later version to point to. (No release in that project since 14-Aug-2014!)

nelz9999:
w00t. 2.10.4 just got released: 
https://github.com/andymccurdy/redis-py/issues/639 
https://github.com/andymccurdy/redis-py/releases

I'll take another go at this soon.

nelz9999:
@jamesls This should be ready to go.
Looks like the Python 3.4 Travis job (https://travis-ci.org/jamesls/fakeredis/jobs/88928129) was having DNS issues:
```fatal: unable to access 'https://github.com/jamesls/fakeredis.git/': Couldn't resolve host 'github.com'```

PauloMigAlmeida:
@jamesls  Any update on this ?

nelz9999:
*bump* This is ready to go out!

jamesls:
Sorry for the delay, I'll take a look tonight.

jamesls:
Thanks!

PR #95, opened 2015-11-09
Improved the build process

thedrow:
Added python 3.5 and pypy to the build and enabled caching to dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,15 @@
 language: python
+sudo: false
+cache:
+  directories:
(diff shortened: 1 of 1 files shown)


thedrow:
It seems that we don't support PyPy yet.
Any idea why this error occurs on PyPy:
https://travis-ci.org/jamesls/fakeredis/jobs/90091266#L220

bmerry:
Looks like pypy works now and I've added it to `.travis.yml`, so closing.

PR #97, opened 2015-12-10
Match redis-py's transaction() and pipeline() functions.

### LINKED ISSUE #78: Behaviour of transaction() does not match redis-py
Here is redis-py's [`transaction()`](https://github.com/andymccurdy/redis-py/blob/c4e2f56dbf6e04a23e6a346ac54278cfbf4035e1/redis/client.py#L473)

As you can see, it takes keyword arguments: `def transaction(self, func, *watches, **kwargs)`
fakeredis' [`transaction()`](https://github.com/jamesls/fakeredis/blob/master/fakeredis.py#L1257) does not expect keyword arguments: `def transaction(self, func, *keys)`

Naturally this results in a `TypeError` when trying to use keyword arguments.

The three current kwargs are `shard_hint=None`, `value_from_callable=False`, and `watch_delay=None`. The fakeredis implementation, however, does not support any of these.

nfvs:
Match the function signatures, add support for `value_from_callable`
and `watch_delay` kwargs. Add test for `value_from_callable`.

Fixes #78.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -1267,7 +1267,7 @@ class FakeStrictRedis(object):
             keys.extend(args)
         return keys
 
-    def pipeline(self, transaction=True):
+    def pipeline(self, transaction=True, shard_hint=None):
         """Return an object that can be used to issue Redis commands in a batch.
 
         Arguments --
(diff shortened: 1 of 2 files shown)


nfvs:
Any feedback?

scarchik:
Hi guys,

@jamesls, can you please give us feedback about this pull request? This is very important change for me.

bmerry:

PR #98, opened 2015-12-17
Implement hscan and sscan methods

nelz9999:
I centralized some common logic between `scan`, `hscan`, and `sscan`. I also added the associated "*_iter" methods: `hscan_iter` and `sscan_iter`.

These methods should be fully API-compatible with redis-py, but they are NOT designed to be performant with large data sets, though they should be fine with smaller testing-sized data.


jamesls:
Thanks!

Issue #107, opened 2016-02-19
fakeredis does not work on Windows because of a libc dependency


krakip:
Currently fakeredis only looks for libc when searching for an implementation of strtod. On Windows, libc is not available and this function is found in msvcrt.lib.

The fix is to check for both libraries and use whichever is present.

jamesls:
Fixed via #108

PR #108, opened 2016-02-19
Look for strtod in msvcrt.lib on Windows; test strtod search behavior

krakip:
This patch fixes issue #107.

I verified that the tests pass on Windows using Python 3.5 and on OS X using Python 2.7 and Python 3.5.


jamesls:
Thanks!

PR #111, opened 2016-04-11
set() ex and px parameters accept timedelta

saabeilin:
The 'true' redis-py library accepts timedeltas for `ex` and `px` parameters; if the expiration is negative, it also throws an exception. 

This pull request fixes these two things to follow redis-py.


jamesls:
Thanks!

PR #105, opened 2016-02-04
Conform new pubsub function api

joac:

PR #114, opened 2016-05-18
Add possibility to ignore subscribe messages when using pubsub

lburg:
Inspired from https://github.com/jamesls/fakeredis/pull/105, I added some tests. Is it enough?


jamesls:
Thanks for the pull request, this has been merged in https://github.com/jamesls/fakeredis/commit/5b6932673ca1c256f9c255540cb36dd435a83be9

PR #82, opened 2015-06-16
Add support for decode_responses

jjackson26:
Implementation for decoding responses as requested in #77.

PR #116, opened 2016-05-31
[rebased] Add support for decode_responses=True

Korijn:
This is a rebased version of #82.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -1,15 +1,15 @@
-import random
-import warnings
+from collections import Iterable, MutableMapping
 import copy
 from ctypes import CDLL, POINTER, c_double, c_char_p, pointer
 from ctypes.util import find_library
-import fnmatch
-from collections import MutableMapping
 from datetime import datetime, timedelta
+import fnmatch
 import operator
+import random
 import sys
 import time
 import re
+import warnings
 
 import redis
 from redis.exceptions import ResponseError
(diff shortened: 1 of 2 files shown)


bersace:
:up: Great contribution ! I need it here too.

jamesls:
Thanks for the pull request, and thanks for rebasing.  My only hesitation with this pull request is that it's fairly invasive and touches just about every method.  I wonder if there's a more succinct way to do this?  My other concern is how likely it's going to be that any new methods/ops added in the future will miss a `_decode_response()` call.  If the tests pass for a new method without requiring a `_decode_response()` call, it seems like this will be something that's easy to miss?

I'm not opposed to this pull request as is, but would like to hear other's thoughts first.

Korijn:
We might be able to move part of the code into `_StrKeyDict`? It won't cover all of it though.

matt-snider:
Hi, two major points where the implementation could be improved:

1. The decode methods are unnecessarily re-declared on every single call to `_decode_response()`. Is there any reason for this?
2. The tests are too simple. They should re-use the existing tests, taking into account `decode_responses`.

I'm currently trying to implement a less invasive method, but it's a bit tricky.

jamesls:
Resolved via https://github.com/jamesls/fakeredis/pull/123 .  Thanks everyone for looking into this.

PR #118, opened 2016-06-01
Check keys exist before delete in _StrKeyDict._update_expired_keys

mmcmahon-jana:
I was running into an issue when use fakeredis and freezegun to manipulate dates in a unit test. The _StrKeyDict._update_expired_keys was trying to delete keys that appeared the _ex_keys dict, but somehow the keys did not appear in the underlying _dict of the class instance.
This just updates the method to verify that the key is contained in the _ex_keys and _dict objects before calling del.

This method is called repeatedly while existing tests are run and I do not want to add a dependency on freezegun to recreate the circumstances of my specific bug.  So I hope that this PR is ok without new tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
   - "2.7"
   - "3.3"
   - "3.4"
+  - "3.5"
 sudo: false
 cache:
   - pip


kairichard:
@mmcmahon-jana are you using django-redis for session by any chance? Because then the problem is related to https://github.com/jamesls/fakeredis/issues/120 as I experienced the same behavior as you did together with freezegun. A potential fix would be to call `to_bytes` when inserting keys into `_ex_keys`.

mmcmahon-jana:
@kairichard interesting, I'm not using django-redis but I will try the to_bytes fix you are suggesting to see if it makes a difference.  Thanks.

bmerry:
I've created a fork called [fakenewsredis](https://github.com/ska-sa/fakenewsredis) with a PyPI [release](https://pypi.python.org/pypi/fakenewsredis) to incorporate PRs that aren't getting traction here. I've included this #128 in fakenewsredis, which I think removes the need for this PR (but let me know if I've misunderstood the issue).

bmerry:
Closing because the structure of the internal storage has changed in fakeredis 0.10.1, which should fix this issue.

Issue #77, opened 2015-05-03
Support the decode_responses argument


twidi:
In `redis-py`, `StrictRedis` accepts a `decode_responses` argument, set to `False` by default, but when set to `True`, it will decode all bytes values using its `charset` argument, default to utf-8.

When my team uses Redis, we set this argument to `True`. So we expect strings (we use python 3), not bytes.
But `fakeredis` always returns bytes. It encodes everything in bytes but never decodes.

So in our tests we have inconsistent behavior if we swap between a real redis server and `fakeredis`.

Viewing the code of `fakeredis` it's not so easy to support it because not everything goes through a single method, as does `redis-py` with `read_response`.

Do you think it's possible to have this feature? Would you accept a pull-request on this subject?

twidi:
By the way, we use a layer on top of redis/fakeredis, so we have a method through which every response go, so I was able to update it this way:

```python
def decode_result(self, result):
    encoding = self.connection._encoding  # FakeStrictRedis

    def dec_bytes(string):
        return string.decode(encoding)

    def dec_iter(val):
        return val.__class__((dec(v) for v in val))

    def dec_items(val):
        return val.__class__(((dec(k), dec(v)) for k, v in val.items()))

    def dec(val):

        # Decode bytes.
        if isinstance(val, bytes):
            return dec_bytes(val)

        # Decode dicts or dict-like objects
        if hasattr(val, 'items'):
            return dec_items(val)

        # Decode other iterable, but not strings.
        if not isinstance(val, str) and isinstance(val, Iterable):
            return dec_iter(val)

        # No decoding for int, float, str
        return val

    return dec(result)
```

Not sure yet if it will handle all the cases, but maybe using something like that in every redis-like functions that return something could help.

matt-snider:
This is something I would also like to use

SpotlightKid:
We would also need this for out tests, since we use `redis.StrictRedis` with `decode_responses` in out production code.

jamesls:
Implemented in #123

PR #123, opened 2016-08-13
Add support for decode_responses

matt-snider:
This PR adds the `decode_responses` functionality requested in #77, and attempted in #116/#82. 

This implementation is an attempt at adding the required logic without needing to change every method/return-statement. I also wanted this change to be well-tested without rewriting code, so I've written the new test classes using the inheritance pattern already used.

This approach is good because it means internally the library can continue to work with only binary data, and things (*should*) just work. I do see some potential downsides though:
 
1. Method patching is considered by some to be too *magical* (although for a library whose main purpose is unit-testing, patching is more acceptable)
2. Inheriting the tests this way means there are now 1161 tests :grin:
3. Since binary literals were used in all existing test cases, it was necessary to override a few `assert*` methods which rely on the redis expression coming as the first parameter, and the literal as the second. This might be overlooked and could cause problems.

I'd appreciate any feedback! :smile:

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -9,6 +9,7 @@ from datetime import datetime, timedelta
 import operator
 import sys
 import time
+import types
 import re
 
 import redis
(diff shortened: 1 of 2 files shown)


jamesls:
Hi, thanks for the pull request, I'm taking a look now.

jamesls:
Looks good to me, and addresses the concerns I had in #116.  Thanks again for the pull request.

matt-snider:
Glad to have been able to help :smile:

ak64th:
在 2016年08月14日 07:18, James Saryerwinnie 写道:

PR #128, opened 2016-09-20
Remove `_ex_keys` mapping

ticosax:
Because they can get out of sync under python3 since
`_dict` entries are converted into bytes while `_ex_keys` aren't
and `b'foo' != 'foo'` under python 3.

Instead of trying to convert back and forth between `bytes` and `str` in different places,
this new implementation stores everything in `_dict` and
 makes sure values and timestamp are always stored together.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -110,14 +110,17 @@ def timedelta_total_seconds(delta):
 class _StrKeyDict(MutableMapping):
     def __init__(self, *args, **kwargs):
         self._dict = dict(*args, **kwargs)
-        self._ex_keys = {}
 
     def __getitem__(self, key):
-        self._update_expired_keys()
-        return self._dict[to_bytes(key)]
+        now = datetime.now()
+        value, expiration = self._dict[to_bytes(key)]
+        if expiration is not None and now > expiration:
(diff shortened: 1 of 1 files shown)


jamesls:
Thanks for the pull request.

Per the [contributing guidelines](https://github.com/jamesls/fakeredis/blob/master/CONTRIBUTING.rst), can you add a test that demonstrates the bug?  That way we can verify this fixes the issue and prevent regressions in the future.  Thanks.

ticosax:
The fix is deprecating any test I could provide since I'm removing `_ex_keys`. and the bug exists because `_ex_keys` exists. To be honest I started by writing a test, try to fix the bug, then realize it was easier to fix by removing `_ex_keys` and the test I wrote didn't make sense anymore.
What do you suggest ?

jdufresne:
I think the test, `test_set_existing_key_persists()`, added in PR #137 would be one that demonstrates the improvement.

blueyed:
@ticosax 
Needs to get rebased also.

ticosax:
thank you @blueyed for the reminder.

bmerry:

PR #130, opened 2016-10-02
Fix bugs related to inconsistent storage of bytes/strings

matt-snider:
This PR fixes a few issues that arise when methods with decoded responses are called within the implementation of a redis command, and then these decoded results are stored without being converted back to bytes.

For example in `sdiffstore()`:
```python
def sdiffstore(self, dest, keys, *args):
    """
    Store the difference of sets specified by ``keys`` into a new
    set named ``dest``.  Returns the number of keys in the new set.
    """
    diff = self.sdiff(keys, *args)
    self._db[dest] = diff  # this line is problematic
    return len(diff)
```

And just to demo this behaviour:
```python
>>> from fakeredis import FakeStrictRedis
>>> r = FakeStrictRedis(decode_responses=True)
>>> r.sadd('foo', '1', '2')
>>> r.sadd('bar', '2', '3')
>>> r.sdiffstore('baz', 'foo', 'bar')

# things look okay at first
>>> r.smembers('baz')
{'1'}
>>> r.scard('baz')
1

# but, the following should have no effect, and does
>>> r.sadd('baz', '1')
>>> r.smembers('baz')
{'1'}
>>> r.scard('baz')
2  # uh oh...

# upon taking a deeper look, the cause is quite obvious:
>>> r._db.__dict__
{'_ex_keys': {}, '_dict': {b'baz': {'1', b'1'}, b'bar': {b'3', b'2'}, b'foo': {b'1', b'2'}}}
```

So yeah, some methods that have decoded responses are being called internally and then their results are saved inconsistently as bytes and strings. To solve this we just need to ensure we use `to_bytes()` properly when storing values as is already done in most methods.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -760,6 +760,7 @@ class FakeStrictRedis(object):
     def rpoplpush(self, src, dst):
         el = self.rpop(src)
         if el is not None:
+            el = to_bytes(el)
             try:
                 self._db[dst].insert(0, el)
             except KeyError:
@@ -793,6 +794,7 @@ class FakeStrictRedis(object):
     def brpoplpush(self, src, dst, timeout=0):
         el = self.rpop(src)
         if el is not None:
+            el = to_bytes(el)
             try:
                 self._db[dst].insert(0, el)
             except KeyError:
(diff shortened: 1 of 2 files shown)


jamesls:
Thanks!

PR #133, opened 2016-11-11
Pass through additional kwargs from `from_url` to the class initializer.

nfvs:
When initializing fake redis with `from_url()`, passing additional parameters like `decode_responses=True` will have no effect, since so far it was ignoring them. This breaks `FakeStrictRedis.from_url()` in python 3.

This PR passes them through to the constructor, like redis-py does:
https://redis-py.readthedocs.io/en/latest/_modules/redis/client.html#StrictRedis.from_url


jamesls:
Thanks!

Issue #101, opened 2015-12-18
Should raise an error when attempting to get a key holding a list


bluemoo:
Here's what happens when I try to use 'get' on a key that I've used to store a list:

```
>>> from redis_cache import get_redis_connection
>>> client = get_redis_connection('persistent')
>>> client.rpush('foo', 1)
1L
>>> client.get('foo')
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/home/noah/envs/adioso/lib/python2.6/site-packages/redis/client.py", line 551, in get
    return self.execute_command('GET', name)
  File "/home/noah/envs/adioso/lib/python2.6/site-packages/redis/client.py", line 361, in execute_command
    return self.parse_response(connection, command_name, **options)
  File "/home/noah/envs/adioso/lib/python2.6/site-packages/redis/client.py", line 371, in parse_response
    response = connection.read_response()
  File "/home/noah/envs/adioso/lib/python2.6/site-packages/redis/connection.py", line 311, in read_response
    raise response
ResponseError: WRONGTYPE Operation against a key holding the wrong kind of value
```

Using fakeredis, I don't get any error:

```
>>> import fakeredis
>>> client = fakeredis.FakeStrictRedis()
>>> client.rpush('foo', 1)
1
>>> client.get('foo')
"['1']"
```

I think the fix to this might involve tracking the type of item at each key, which could get involved. Is this of interest to fakeredis?

bmerry:
I'm currently working on a more complete version of this, which will do type checks in all commands.

PR #136, opened 2017-01-07
Fix Issue #101 - Should raise an error when attempting to get a key holding a list 

TereFDS:
It fixes Issue  https://github.com/jamesls/fakeredis/issues/101 
This bug also extends to Set and SortedSet structures (and not only list)
It updates `FakeRedis.get` method in order to raise the error while doing get from a list.

@jamesls

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -16,6 +16,8 @@ import redis
 from redis.exceptions import ResponseError
 import redis.client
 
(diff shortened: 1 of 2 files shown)


bmerry:

bmerry:
Closing because #166 is included in fakeredis 0.10.1.

PR #138, opened 2017-01-28
Add Python 3.6 to Travis CI and document in setup.py

jdufresne:


jdufresne:
> maybe add nightly also ?

Good idea. Done in latest revision.

Issue #91, opened 2015-08-24
Unicode errors in set()


domenkozar:
```
feeds/cache.py:128: in store_feed_content
    self.redis_client.set(feed_key, raw_content)
.tox/py27/lib/python2.7/site-packages/fakeredis.py:386: in set
    self._db[name] = to_bytes(value)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

x = '<feedán/>', charset = 'ascii', errors = 'strict'

    def to_bytes(x, charset=sys.getdefaultencoding(), errors='strict'):
        if x is None:
            return None
        if isinstance(x, (bytes, bytearray, buffer)) or hasattr(x, '__str__'):
>           return bytes(x)
E           UnicodeEncodeError: 'ascii' codec can't encode character u'\xe1' in position 5: ordinal not in range(128)
```

danielgoncalves:
Indeed. When you use set using **redis-py**:

```python
>>> import fakeredis
>>> r = fakeredis.FakeStrictRedis()
>>> r.set('something', u'Alçapão')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/daniel/.virtualenvs/ibpt/local/lib/python2.7/site-packages/fakeredis.py", line 386, in set
    self._db[name] = to_bytes(value)
  File "/home/daniel/.virtualenvs/ibpt/local/lib/python2.7/site-packages/fakeredis.py", line 31, in to_bytes
    return bytes(x)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe7' in position 2: ordinal not in range(128)
```

Using **redis-py**:

```python
>>> import redis
>>> r = redis.StrictRedis()
>>> r.set('something', u'Alçapão')
True
```

vrcmarcos:
I'm facing this problem right now. Any solution?

TereFDS:
[...]

PR #142, opened 2017-02-07
Fix Issue #91 - Unicode errors in set

TereFDS:
Fixes: https://github.com/jamesls/fakeredis/issues/91


jamesls:
Fixed via https://github.com/jamesls/fakeredis/pull/160

PR #148, opened 2017-04-25
Speed up listen loop

robodan:
Read queued messages as fast as possible (no delay)
When the queue is empty, delay by 0.1 sec (was 1 sec) before checking again

I didn't add a test because there are no logic changes and timing tests tend to be flaky.
This speeds up the test suite by about 20sec.


jamesls:
Thanks!

PR #149, opened 2017-05-15
type('a_bad_key') returns the string 'none', not None

catwell:
This mimics the behavior of `redis-py`.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -547,6 +547,9 @@ class FakeStrictRedis(object):
             return b'list'
         elif isinstance(key, set):
             return b'set'
+        else:
+            assert key is None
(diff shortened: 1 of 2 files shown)


fakeredis.py
```diff
@@ -547,6 +547,9 @@ def type(self, name):
             return b'list'
         elif isinstance(key, set):
             return b'set'
+        else:
+            assert key is None
```
jamesls: Should the assert be here?  Is it possible for key to be not be None at this point?

catwell: If the key is not `None`, it means the code is missing a case and the return value of the function will be wrong anyway (it will return `None` instead of a string).

PR #150, opened 2017-05-23
Support b'bytes' keys in `sinter` and similar.

amw:
redis-py allows you to pass `bytes` objects as keys. There are many places where fakeredis does not support that and this patch does not fix all of them, but it at least fixes `sinter` and other methods that use `_list_or_args`.


jamesls:
Thanks for the pull request!

Issue #132, opened 2016-10-19
srandmember does not support `number` argument


RossLarner:
Thanks for the great library! It would be fantastic if fakeredis supported the optional `number` argument to `srandmember`, returning a list of `number` random members of the given set (available in Redis 2.6+)

PR #152, opened 2017-06-22
Support `number` argument in `srandmember`

amw:
Related to #132


amw:
I have not noticed the Travis failure. Bummer.

Good that it's supported one way or another.

PR #154, opened 2017-07-03
Make the fallback value bytes in getbit()

dsalisbury:
Make the fallback value bytes in `getbit()` otherwise `byte_to_int()` will go bang if `getbit(key, n)` is called before `setbit(key, n, v)`

Specifically, I have a test which fails due to this:

```
b = '\x00'

    def byte_to_int(b):
        if isinstance(b, int):
            return b
>       raise TypeError('an integer is required')
E       TypeError: an integer is required
```

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -323,7 +323,7 @@ class FakeStrictRedis(object):
 
     def getbit(self, name, offset):
         """Returns a boolean indicating the value of ``offset`` in ``name``"""
-        val = self._db.get(name, '\x00')
+        val = self._db.get(name, b'\x00')
         byte = offset // 8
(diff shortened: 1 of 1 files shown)


bmerry:
This is also handled by #166.

bmerry:
I've created a fork called [fakenewsredis](https://github.com/ska-sa/fakenewsredis) with a PyPI [release](https://pypi.python.org/pypi/fakenewsredis) to incorporate PRs that aren't getting traction here. I've included #166 (which supersedes this one) there.

bmerry:
Closing because #166 has been merged into 0.10.1 (via the fakenewsredis fork).

PR #156, opened 2017-07-11
srandmember number argument and tests.

Tinche:
Like #152, except it should work on all supported Pythons.


jamesls:
Thanks!

Tinche:
My pleasure ;)

PR #160, opened 2017-10-02
Fix issue #91

### LINKED ISSUE #91: Unicode errors in set()

faustow:
Forces encoding to UTF-8 when using python 2.x
Fixes #91

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -40,7 +40,10 @@ if PY2:
 
     def to_bytes(x, charset=DEFAULT_ENCODING, errors='strict'):
         if isinstance(x, (bytes, bytearray, buffer)) or hasattr(x, '__str__'):
-            return bytes(x.encode(charset, errors))
+	    try:
+                return bytes(x)
+	    except Exception:
+                return bytes(x.encode(charset, errors))
         if isinstance(x, unicode):
             return x.encode(charset, errors)
         if hasattr(x, '__unicode__'):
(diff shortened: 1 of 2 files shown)


fakeredis.py
```diff
@@ -25,23 +26,24 @@
...
-if not PY2:
-    long = int
-
 
 __version__ = '0.8.2'
 
 
-if sys.version_info[0] == 2:
+if PY2:
+    DEFAULT_ENCODING = 'utf-8'
```
jamesls: Only defining `DEFAULT_ENCODING` for py2 means you have to be really careful about never using `DEFAULT_ENCODING` in py3.  The intent here is to have all the constants defined for py2 as well as py3.

jamesls:
Merged via f8c3921

PR #161, opened 2017-10-07
Implement SETRANGE command

### LINKED ISSUE #157: setrange not implemented
Hi,

Just a note that the setrange is not implemented (just a pass)

Could possibly be implemented with:

```
 setrange(self, name, offset, value):
     _tmp = self.get(name)
     if not _tmp:
         _tmp = ""
     _tmp = _tmp[0:offset] + value + _tmp[offset+len(value):]
     self.set(name, _tmp)
```

Matoking:
Fixes #157 

Implements SETRANGE command and a few test cases for the command.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -503,7 +503,12 @@ class FakeStrictRedis(object):
         return result
 
     def setrange(self, name, offset, value):
-        pass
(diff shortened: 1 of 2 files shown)


jamesls:
Thanks!

Issue #165, opened 2017-11-09
linsert implementation is incomplete


bmerry:
It
- does not examine `where` (it only implements `BEFORE`)
- does not handle the case of the pivot not being in the list.

I have a fix in a branch I am preparing for PR.

bmerry:
Fixed in 0.10.1.

Issue #164, opened 2017-11-09
Several incr/decr functions do not convert the value to bytes


bmerry:
The implementations of `decr`, `hincrby` and `hincrbyfloat` store the int/float value internally, instead of the stringified version. This can lead to errors e.g.
```python
>>> f = fakeredis.FakeStrictRedis()
>>> f.decr('key', 1)
-1
>>> f.append('key', 'x')
TypeError: unsupported operand type(s) for +=: 'int' and 'str'
```
I have a fix in a branch I am preparing for PR.

bmerry:
This is a duplicate of #90, so I'm closing it.

Issue #163, opened 2017-11-09
getset does not to_bytes the value


bmerry:
In 0.9.0, the implementation of `getset` doesn't pass the value through `to_bytes` (unlike `set`).

This leads to problems like the following:
```python
>>> f = fakeredis.FakeStrictRedis()
>>> f.getset('key', 3)
>>> f.append('key', 'hello')
TypeError: unsupported operand type(s) for +=: 'int' and 'str'
```

I have a fix in a branch I am preparing for a PR.

bmerry:
Fixed in 0.10.1.

PR #166, opened 2017-11-09
Add type checking

### LINKED ISSUE #101: Should raise an error when attempting to get a key holding a list

bmerry:
This extends all the commands that operate on existing data of a specific type to raise a `redis.ResponseError` if the key exists but is not of that type. This fixes #101, but is far more general than #136.

The stricter checking exposed some bugs, so this PR also contains fixes (with unit tests) for #163, #164, #165.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -249,9 +249,22 @@ class FakeStrictRedis(object):
 
         del self._pubsubs[:]
 
+    def _get_string(self, name, default=b''):
+        value = self._db.get(name, default)
+        # Allow None so that default can be set as None
+        if not isinstance(value, bytes) and value is not None:
+            raise redis.ResponseError(_WRONGTYPE_MSG)
+        return value
+
+    def _setdefault_string(self, name):
+        value = self._db.setdefault(name, b'')
+        if not isinstance(value, bytes):
+            raise redis.ResponseError(_WRONGTYPE_MSG)
+        return value
+
     # Basic key commands
     def append(self, key, value):
-        self._db.setdefault(key, b'')
+        self._setdefault_string(key)
         self._db[key] += to_bytes(value)
         return len(self._db[key])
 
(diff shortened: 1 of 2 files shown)


bmerry:
@jamesls I know there is a backlog of PRs, but it would be great if this could be looked at before too much other code gets changed: it touches the majority of the functions in the API and so will be nightmare to merge if it diverges too far from master.

bmerry:
One area that isn't covered is distinguishing sets and hyperloglogs. That would take quite a bit more work, because the hyperloglog functions are mapped to set functions.

PR #169, opened 2017-11-20
fix setbit

cosven:
1. setbit SHOULD return old_value instead of None
2. there exists bug when setbit to 0

for example:

```
# originally
self.redis.setbit('foo', 4, 0)
value = self.redis.getbit('foo', 4)  # we will get 1, but we should get 0
```

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -475,13 +475,16 @@ class FakeStrictRedis(object):
             # bit.
             needed = byte - (len(val) - 1)
             val += b'\x00' * needed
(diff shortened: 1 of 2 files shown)


adamantike:
I have experienced this same issue. The simplest scenario where current implementation fails is:

```
def test_getbit_zero(self):
    self.redis.setbit('foo', 1, 0)
    self.assertEqual(self.redis.getbit('foo', 1), 0)
```

@jamesls can you please consider merging this fix?

zucaritask:
Could you release a patch version now that this fix is merged?

PR #172, opened 2018-01-13
Add score_cast_func support

guigen:
Allows to pass a function to cast the score return value.


test_fakeredis.py
```diff
@@ -1368,6 +1368,14 @@ def test_zrangebyscore_withscores(self):
         self.assertEqual(self.redis.zrangebyscore('foo', 1, 3, 0, 2, True),
                          [(b'one', 1), (b'two', 2)])
 
+    def test_zrangebyscore_withscores_score_cast_func(self):
+        self.redis.zadd('foo', one=1)
+        self.redis.zadd('foo', two=2)
+        self.redis.zadd('foo', three=3)
+        self.assertEqual(self.redis.zrangebyscore('foo', 1, 3, 0, 2, True,
+                                                  int),
+                         [(b'one', 1), (b'two', 2)])
```
bmerry: This is not an effective test, because it passes even if the score_cast_func isn't being applied. Rather use `str` as the cast func.

test_fakeredis.py
```diff
@@ -1368,6 +1368,14 @@ def test_zrangebyscore_withscores(self):
...
+
     def test_zrevrangebyscore(self):
```
bmerry: Add a test for zrevrangebyscore with score_cast_func as well.

PR #174, opened 2018-01-19
Fix getbit for initial value of 0 for python3

zucaritask:
When calling the redis.getbit operation with any offset, the initial
value could not be calculated due to a TypeError in the byte_to_int
operation expecting an int.

This commit fixes the getbit command for its initial value by making it
a byte literal


bmerry:
I'm closing this because I believe it is fixed (in a different way) in fakeredis 0.10.1. If that's not the case, please can you rebase against master and re-submit.

PR #175, opened 2018-02-01
Support py-redis 2.10.6

kouk:
In 2.10.6 the SET command will pass a zero expiration directly to the
server, which returns an error. Previously it ignored it.

https://github.com/andymccurdy/redis-py/blob/d48c46d3f8c6943c7995bffbecb0fb4c66daabe2/CHANGES#L8


bmerry:
Closing because it's been independently fixed (as part of the fakenewsredis fork, which has been merged into 0.10.1).

Issue #176, opened 2018-02-02
AttributeError: 'FakeStrictRedis' object has no attribute 'eval'


Panter7777:
fakeredis==0.9.0
Django==1.11.5
redis==2.10.6
django-redis==4.8.0

use cache from django `from django.core.cache import cache`
```python 
# settings.py
CACHES = {
    'default': {
        'BACKEND': 'django_redis.cache.RedisCache',
        'LOCATION': REDIS_URL,
        'OPTIONS': {
            'CLIENT_CLASS': 'django_redis.client.DefaultClient',
            'SERIALIZER': 'django_redis.serializers.json.JSONSerializer',
        },
        'KEY_PREFIX': 'backend',
    }
}
# Settings for tests
if 'test' in sys.argv[1:]:
    import fakeredis
    CACHES['default']['OPTIONS']['REDIS_CLIENT_CLASS'] = "fakeredis.FakeStrictRedis"

````
```python
File ".../customer/views.py", line 111, in login
    cache.incr(key)
  File ".../lib/python3.5/site-packages/django_redis/cache.py", line 33, in _decorator
    return method(self, *args, **kwargs)
  File ".../lib/python3.5/site-packages/django_redis/cache.py", line 117, in incr
    return self.client.incr(*args, **kwargs)
  File ".../lib/python3.5/site-packages/django_redis/client/default.py", line 427, in incr
    return self._incr(key=key, delta=delta, version=version, client=client)
  File ".../lib/python3.5/site-packages/django_redis/client/default.py", line 398, in _incr
    value = client.eval(lua, 1, key, delta)
AttributeError: 'FakeStrictRedis' object has no attribute 'eval'
```

Panter7777:
When you going to create new release with this fix?

bmerry:
Probably next week, although if I get more PRs to improve the scripting support further then I'll try to get those in first.

Panter7777:
Hello. 25 days passed, what about fixes? When you going to create new release?

bmerry:
It's in the [fakenewsredis](https://github.com/ska-sa/fakenewsredis) 0.10.0 release. At some point in the future I want to merge the fakenewsredis fork back into fakeredis (I've recently been given the access to do so, but won't have time this month), so in the meantime I suggest using fakenewsredis (and filing any issues with it there).

PR #177, opened 2018-02-05
Support StrictRedis.eval for Lua scripts

blfoster:
Add support for Lua scripts with ScrictRedis.eval. Does not support register_script, although adding it should be simple from here. Fixes https://github.com/jamesls/fakeredis/issues/176


.travis.yml
```diff
@@ -6,7 +6,6 @@ python:
   - 3.4
   - 3.5
   - 3.6
-  - nightly
```
blfoster: Looking at the build history, this seems to be flaky. I'm not sure that there's much value in keeping it.

blfoster:
This has been merged in https://github.com/ska-sa/fakenewsredis.

PR #184, opened 2018-04-02
Replace fnmatch with hand-coded pattern compiler

### LINKED ISSUE #182: Pattern matching does not match redis behaviour
The implementation of KEYS and SCAN uses `fnmatch.fnmatch` to match the pattern. This has a few problems:
- At a minimum it should use fnmatchcase to avoid being case-insensitive on some OSes.
- It does not support backslash escapes.
- It uses `!` instead of `^` for negated
- It seems to handle invalid ranges like `[e-a]` differently (raises an exception, whereas redis does something unspecified I still need to work out).

It also operates by turning both arguments into native strings using the system default encoding, even though they might not be valid in that encoding.

bmerry:
This matches the quirks in the redis pattern matching, which has some
differences from fnmatch. It also applies all the logic in bytes, rather
than the native character encoding.

Fixes #182.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -2002,7 +2002,7 @@ class FakeStrictRedis(object):
         else:
             regex = None
         for val in data[cursor:result_cursor]:
-            if not regex or regex.match(val):
+            if not regex or regex.match(to_bytes(val)):
                 result_data.append(val)
         if result_cursor >= len(data):
             result_cursor = 0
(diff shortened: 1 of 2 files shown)


bmerry:
I see that the pubsub mechanism has its own (incorrect) pattern compiler, so I'll try to get that converted too before merging. It's complicated by the pubsub code not correctly dealing with str vs bytes.

PR #189, opened 2018-04-25
Add 'System' to the list of libc equivalents

### LINKED ISSUE #188: Import fails on osx, says it cannot find libc or msvcrt
libc.dylib exists but ctypes skips it. Possibly because it is a symlink.

I added `find_library('System')` which is what libc links to and the import worked.

bmerry:
This is apparently needed on some OSX systems. Closes #188.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -88,7 +88,7 @@ else:
 
 DATABASES = {}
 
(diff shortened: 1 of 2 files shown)


shaleh:
:+1:

PR #192, opened 2018-05-22
Make __getitem__ raise a KeyError for missing keys

### LINKED ISSUE #191: __getitem__ behavior does not match redis-py
Environment:
- Python 3.6.5
- fakeredis-0.10.3
- redis-2.10.6

Example:
```
>>> from redis import StrictRedis
>>> from fakeredis import FakeStrictRedis
>>> 
>>> redis = StrictRedis(host='localhost', port=6379, decode_responses=True)
>>> fakeredis = FakeStrictRedis(decode_responses=True)
>>> redis['non-existent key']
  File "<stdin>", line 1, in <module>
  File "/home/dnoetzel/.local/share/virtualenvs/dnoetzel-8_UrYaNn/lib/python3.6/site-packages/redis/client.py", line 986, in __getitem__
    raise KeyError(name)
KeyError: 'non-existent key'
>>> print(fakeredis['non-existent key'])
None
>>> 
```

Expected behavior:
`__getitem__` raises a `KeyError` when trying to fetch a non-existent key (as redis-py [does](https://github.com/andymccurdy/redis-py/blob/2035e9821ebf2d884f10928e9c88e17179ac4842/redis/client.py#L979))
Actual behavior:
`__getitem__` returns `None`

dnoetzel:
In redis-py, attempting to use __getitem__ with a non-existent key
causes a KeyError to be raised, while fakeredis returns None. This
change makes it so that __getitem__ in fakeredis works the same way as
in redis-py.

Fixes #191

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -469,7 +469,10 @@ class FakeStrictRedis(object):
             return to_bytes(value)
 
     def __getitem__(self, name):
-        return self.get(name)
+        value = self.get(name)
+        if value is not None:
+            return value
+        raise KeyError(name)
 
     def getbit(self, name, offset):
         """Returns a boolean indicating the value of ``offset`` in ``name``"""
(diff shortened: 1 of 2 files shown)


bmerry:
Thanks, looks good!

Issue #44, opened 2014-03-26
Support the score_cast_func in redis 2.9.x


mbox:
The latest version of redis-py has a "score_cast_func" parameter on the sorted set functions that accept a withscores parameter.

As fakeredis doesn't have a matching parameter, attempts to use with such calls fail

bmerry:
Implemented in #194 and included in 0.11.0 release.

PR #194, opened 2018-06-08
Supports argument score_cast_func in zrangebyscore()

### LINKED ISSUE #193: Unsupported keyword argument 'score_cast_func' to zrangebyscore()
In the latest version of `redis-py`, [zrangebyscore() accepts a keyword argument called score_cast_func.](http://redis-py.readthedocs.io/en/latest/#redis.StrictRedis.zrangebyscore)

Fakeredis [does not support it yet](https://github.com/jamesls/fakeredis/blob/master/fakeredis.py#L1630-L1631).

I'll do a PR ASAP.

fferrara:
Fixes #193  #44

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -1593,7 +1593,7 @@ class FakeStrictRedis(object):
                                 lambda x: x in
                                 valid_keys)
 
-    def zrange(self, name, start, end, desc=False, withscores=False):
+    def zrange(self, name, start, end, desc=False, withscores=False, score_cast_func=float):
         """
         Return a range of values from sorted set ``name`` between
         ``start`` and ``end`` sorted in ascending order.
@@ -1604,6 +1604,8 @@ class FakeStrictRedis(object):
 
         ``withscores`` indicates to return the scores along with the values.
         The return type is a list of (value, score) pairs
+
+        ``score_cast_func`` a callable used to cast the score return value
         """
         if end == -1:
             end = None
@@ -1619,7 +1621,7 @@ class FakeStrictRedis(object):
         if not withscores:
             return items
         else:
-            return [(k, all_items[k]) for k in items]
+            return [(k, score_cast_func(all_items[k])) for k in items]
 
     def _get_zelements_in_order(self, all_items, reverse=False):
         by_keyname = sorted(
@@ -1639,7 +1641,7 @@ class FakeStrictRedis(object):
         ``withscores`` indicates to return the scores along with the values.
         The return type is a list of (value, score) pairs
 
-        `score_cast_func`` a callable used to cast the score return value
+        ``score_cast_func`` a callable used to cast the score return value
         """
         return self._zrangebyscore(name, min, max, start, num, withscores, score_cast_func,
                                    reverse=False)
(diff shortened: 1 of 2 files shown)


fakeredis.py
```diff
@@ -1638,11 +1638,13 @@ def zrangebyscore(self, name, min, max,
 
         ``withscores`` indicates to return the scores along with the values.
         The return type is a list of (value, score) pairs
+
+        `score_cast_func`` a callable used to cast the score return value
```
bmerry: Missing a leading backtick (looks like that's inherited from redis-py, but might as well fix it). Same issue on zrevrangebyscore.

test_fakeredis.py
```diff
@@ -1879,6 +1879,24 @@ def test_zrangebyscore_withscores(self):
...
+        expected_with_cast_round = [(b'two', 2.0), (b'two_a_also', 2.0)]
+        self.assertItemsEqual(
+            self.redis.zrangebyscore('foo', 2, 3, withscores=True),
+            expected_without_cast_round
+        )
+        self.assertItemsEqual(
+            self.redis.zrangebyscore('foo', 2, 3, withscores=True, score_cast_func=round_str),
+            expected_with_cast_round
+        )
+
```
bmerry: Test looks fine, but can you add a similar test for zrevrangebyscore for completeness.

fferrara:
I'll try to add support for `score_cast_func` on the other functions. AFAIK they are `zscan()`, `zscan_iter()`, `zrange()`, `zrevrange()`. Right?

bmerry:
That sounds about right - but I'd suggest just searching for it in redis-py.

bmerry:
Seems like those tests don't pass on real redis. I haven't investigated, but my guess is that redis-py calls the score_cast_func with the raw bytes from the wire, not a float - so you should probably do the same, and then write the tests with a score_cast_func that expects a string (or I'd guess `bytes` in Python 3).

fferrara:
@bmerry I fixes the tests, they must support strings/bytes indeed. How do you feel about this PR?

bmerry:
It's looking good, but the float/str issue doesn't just affect tests, it needs to be reflected in fakeredis.py too: the float score needs to be encoded as a string before being passed to score_cast_func; and it would be nice if the score_cast_func used in the tests verified that they receive a string (I'd also like to just have one copy of the function, instead of a local copy in each test).

bmerry:
In redis-py it is a bytes object, but as far as I can see in fakeredis it is still a float. So to emulate redis-py, fakeredis will need to turn the float it has into a bytes object before passing it to score_cast_func (which in the default case will turn it back into a float again).

fferrara:
@bmerry  I tried to implement what you intended. Is that right?

bmerry:
There's one very minor tweak I'd like to make in the test, but I'll just merge and make the tweak myself. Thanks for the work on this - it's been a missing feature for quite a while.

fferrara:
Thank you for the support! How long do you think until we have a new tag/version?

bmerry:
I should have time to cut a new version some time in the next week.

fferrara:
@bmerry can we have a release for that? I'd like to clean up the weird hacks to work around this error

Issue #102, opened 2015-12-22
Ability to mock no connection to server


Amertz08:
I think it would be useful to mock the errors thrown when you are unable to connect to the redis server. It appears that redis-py throws a `ConnectionError` with these args `('Error 61 connecting to localhost:6379. Connection refused.',)` when it cannot reach the server.

jamesls:
Seems reasonable.  Is someone willing to propose what the API for this would look like?

Amertz08:
```python
class FakeStrictRedis(object):
    CONNECTED = TRUE

    def set(*args):
        if not self.CONNECTED:
            raise ConnectionError('Error 61 connecting to localhost:6379. Connection refused.')
        else:
            # proceed as normal
```

Just an example I guess. Would need to look at the entire redis-py lib

Amertz08:
2 years after the fact I've decided I'd give this a shot. [Here](https://github.com/jamesls/fakeredis/compare/master...Amertz08:102-mock-connection-error) is what I have so far

PR #197, opened 2018-07-07
Mock connection error

Amertz08:
In reference to #102 

Created a decorator to be added to top level methods that will throw a `ConnectionError` when `connect=False` is passed in constructor.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -440,13 +440,14 @@ class FakeStrictRedis(object):
         return name in self._db
     __contains__ = exists
 
+    @check_conn
     def expire(self, name, time):
         return self._expire(name, time)
 
+    @check_conn
     def pexpire(self, name, millis):
         return self._expire(name, millis, 1000)
 
-    @check_conn
     def _expire(self, name, time, multiplier=1):
         if isinstance(time, timedelta):
             time = int(timedelta_total_seconds(time) * multiplier)
@@ -460,13 +461,14 @@ class FakeStrictRedis(object):
         else:
             return False
 
+    @check_conn
     def expireat(self, name, when):
         return self._expireat(name, when)
 
+    @check_conn
     def pexpireat(self, name, when):
         return self._expireat(name, when, 1000)
 
-    @check_conn
     def _expireat(self, name, when, multiplier=1):
         if not isinstance(when, datetime):
             when = datetime.fromtimestamp(when / float(multiplier))
@@ -719,13 +721,14 @@ class FakeStrictRedis(object):
     # according to the docs.
     getrange = substr
 
+    @check_conn
     def ttl(self, name):
         return self._ttl(name)
 
+    @check_conn
     def pttl(self, name):
         return self._ttl(name, 1000)
 
-    @check_conn
     def _ttl(self, name, multiplier=1):
         if name not in self._db:
             return -2
@@ -757,10 +760,12 @@ class FakeStrictRedis(object):
             assert key is None
             return b'none'
 
+    @check_conn
     @_lua_reply(_lua_bool_ok)
     def watch(self, *names):
         pass
 
+    @check_conn
     @_lua_reply(_lua_bool_ok)
     def unwatch(self):
         pass
@@ -827,6 +832,7 @@ class FakeStrictRedis(object):
         except KeyError:
             return []
 
+    @check_conn
     def eval(self, script, numkeys, *keys_and_args):
         from lupa import LuaRuntime, LuaError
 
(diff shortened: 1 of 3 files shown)


fakeredis.py
```diff
@@ -331,6 +331,17 @@ def release(self):
         self.redis.delete(self.name)
 
 
+def check_conn(func):
```
bmerry: I see this an internal function, so should have an underscore prefix.

Amertz08:
@bmerry Yeah I realized working through it that a decorator on every function was becoming a mentally laborious and I had wished there there was a sole underlying function that everything called so I could simply put the logic in it like the actual Redis implementation. I had not noticed the technique used for `_patch_response` and that does look like it might be the best solution to reduce the usage of the decorator.

bmerry:
Sure, that sounds sensible.

fakeredis.py
```diff
@@ -367,18 +366,19 @@ def __init__(self, db=0, charset='utf-8', errors='strict',
         self._encoding_errors = errors
         self._pubsubs = []
         self._decode_responses = decode_responses
-        self._connected = connected
+        self.connected = connected
         if decode_responses:
-            _patch_responses(self)
+            _patch_responses(self, _make_decode_func)
+
+        if not connected:
```
bmerry: Should this be done unconditionally? Otherwise you get the odd behaviour that you can toggle `connected` on and off, but only if it was initially set to false.

fakeredis.py
```diff
@@ -331,12 +331,11 @@ def release(self):
...
 
-def check_conn(func):
+def _check_conn(func):
     """Used to mock connection errors"""
-
     @functools.wraps(func)
     def func_wrapper(*args, **kwargs):
-        if not args[0]._connected:
+        if not func.__self__.connected:
             raise redis.ConnectionError
```
bmerry: Can you also pass a message with the ConnectionError e.g. "fakeredis is emulating a connection error".

bmerry:
Okay, we can leave the tests as is for now - trying to do it by introspection sounds yucky. We can always remove them later if they become annoying to maintain.

Can you perhaps write a section in the README to describe this feature?

fakeredis.py
Amertz08: Yes I believe you are correct. I'll implement that change.

fakeredis.py
Amertz08: Will do. Kinda been thinking that needed a message.

Amertz08:
@bmerry Nix that question on the documentation. I was able to resolve it.

bmerry:
Thanks for all the updates. I'm going to be away for about a week, so I'll take a fresh look when I'm back.

Amertz08:
Yeah I'll have to look over `redis-py` and see how it is implemented and get that in if needed.

Amertz08:
[Here](https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L2408) is the underlying function on the `PubSub` class. Basically the same implementation as the standard Redis class.

bmerry:
> Here is the underlying function on the PubSub class. Basically the same implementation as the standard Redis class. So maybe we implement it in the same manner as we did with FakeStrictRedis

Sounds reasonable to me.

fakeredis.py
```diff
@@ -2231,11 +2231,13 @@ class FakePubSub(object):
     PATTERN_MESSAGE_TYPES = ['psubscribe', 'punsubscribe']
     LISTEN_DELAY = 0.1          # delay between listen loops (seconds)
 
-    def __init__(self, decode_responses=False, *args, **kwargs):
+    def __init__(self, connected=True, decode_responses=False, *args, **kwargs):
```
bmerry: I would rather put `connected` after `decode_responses`, so that it doesn't break existing code that passes decode_responses as a positional parameter.

test_fakeredis.py
```diff
@@ -4403,5 +4403,42 @@ def test_hscan_iter(self):
             self.redis.hscan_iter('name')
 
 
+class TestPubSubConnected(unittest.TestCase):
+
+    def setUp(self):
+        self.redis = fakeredis.FakePubSub(connected=False)
```
bmerry: I would rather call this variable `self.pubsub` since it is not an instance of a Redis object.

test_fakeredis.py
```diff
@@ -4403,5 +4403,42 @@ def test_hscan_iter(self):
             self.redis.hscan_iter('name')
 
 
+class TestPubSubConnected(unittest.TestCase):
+
```
bmerry: For consistency with other classes, remove this blank line.

Amertz08:
Alright sweet. First OSS contrib.

Issue #200, opened 2018-08-09
AttributeError from mock in _check_conn for patched FakeStrictRedis.eval


blueyed:
Because of ed56f4a I cannot patch `eval` anymore easily:

```python
@pytest.fixture
def mocked_redis_eval(mocker):
    from app.authentication import failed_logins_cache

    def mocked_eval(script, numkeys, *keys_and_args):
        """Mock fakeredis' eval (used by django-redis for incr)."""
        assert 'local exists =' in script
        key = keys_and_args[0]
        cur = failed_logins_cache.get(key, 0)
        new = cur + 1
        failed_logins_cache.set(key, new)
        return new
    return mocker.patch('fakeredis.FakeStrictRedis.eval',
                        side_effect=mocked_eval)
```

My main intention here is to not require "lupa" for tests, since only "incr" is used.

The problem is that "mock" raises an AttributeError when magic names are accessed:

```
  File "…/project/.venv/lib/python3.6/site-packages/django_redis/client/default.py", line 431, in incr                                                           
    return self._incr(key=key, delta=delta, version=version, client=client, ignore_key_check=ignore_key_check)                                                                               
  File "…/project/.venv/lib/python3.6/site-packages/django_redis/client/default.py", line 401, in _incr                                                          
    value = client.eval(lua, 1, key, delta)
  File "…/Vcs/fakeredis/fakeredis.py", line 341, in func_wrapper
    if not func.__self__.connected:
  File "…/pyenv/3.6.6/lib/python3.6/unittest/mock.py", line 584, in __getattr__                                                                                         
    raise AttributeError(name)
AttributeError: __self__
```

It gets accessed here: https://github.com/jamesls/fakeredis/blob/b354e472f22551907deccb2bbbd25594f80b24bc/fakeredis.py#L336-L343

Need to investigate myself, but wanted to file this here already.

bmerry:
I think it might work just to use `args[0]` in place of `func.__self__`. If it works I'll happily accept a PR for that.

bmerry:
I think [this documentation](https://docs.python.org/3/library/unittest.mock-examples.html#mocking-unbound-methods) describes an approach that should solve your problem.

Another option if you're not making assertions about the calls to `eval` (just using `mock` to do monkey-patching) is to use
```python
return mocker.patch('fakeredis.FakeStrictRedis.eval', mocked_eval)
```
instead of
```python
return mocker.patch('fakeredis.FakeStrictRedis.eval',
[...]

PR #201, opened 2018-08-09
_check_conn: handle AttributeError with mocked func

blueyed:
Fixes https://github.com/jamesls/fakeredis/issues/200.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -337,7 +337,11 @@ def _check_conn(func):
     """Used to mock connection errors"""
     @functools.wraps(func)
     def func_wrapper(*args, **kwargs):
-        if not func.__self__.connected:
+        try:
+            connected = func.__self__.connected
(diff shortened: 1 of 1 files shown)


bmerry:
I'm not a big fan of this - I feel like this is adding more magic to fix the problems of the initial magic. I'll see if I can come up with something that solves your problem with less magic instead of more.

blueyed:
Agreed.
I will try your suggestions in #200, and maybe something like this is not needed than at all?
(but it certainly was confusing when trying to upgrade fakeredis with my current mocking approach)

bmerry:
Yes, if you can fix this from the outside using my suggestions in #200 then I'd be happiest to make no changes at all.

blueyed:
autospec=True works: https://github.com/jamesls/fakeredis/issue_comments#issuecomment-412081913

PR #202, opened 2018-08-12
Function smembers returns deepcopy

lowitea:


lowitea:
@bmerry done

bmerry:
I think that's not the best way to test this. As far as I'm aware Python doesn't guarantee that mutating a sequence while iterating over it will lead to an exception, and even if it did, it's an indirect way of testing that the value returned by smembers is not affected by future redis operations on the set. A more direct test would be to
1. Add some items to a set with `sadd`
2. Get the set with `smembers` and store it in a variable
3. Add some more items to the set
4. Check that the set stored in the variable did not change.

bmerry:
Thanks.

Issue #204, opened 2018-08-19
Can't Iterate over ARGV or HGETALL results in Lua scripting


NimrodParasol:
When using `eval` to run Lua script, when iterating over ARGV using the following code:
```
for i, v in ipairs(ARGV) do
end
```
An exception is thrown: 

```
/home/ubuntu/venv3/lib/python3.5/site-packages/fakeredis.py:2201: in execute
    for name, args, kwargs in self.commands]
/home/ubuntu/venv3/lib/python3.5/site-packages/fakeredis.py:2201: in <listcomp>
    for name, args, kwargs in self.commands]
/home/ubuntu/venv3/lib/python3.5/site-packages/fakeredis.py:221: in decode_response
    val = _decode(func(*args, **kwargs))
/home/ubuntu/venv3/lib/python3.5/site-packages/fakeredis.py:342: in func_wrapper
    return func(*args, **kwargs)
/home/ubuntu/venv3/lib/python3.5/site-packages/fakeredis.py:849: in eval
    result = lua_runtime.execute(script)
lupa/_lupa.pyx:277: in lupa._lupa.LuaRuntime.execute
    ???
lupa/_lupa.pyx:1294: in lupa._lupa.run_lua
    ???
lupa/_lupa.pyx:1303: in lupa._lupa.call_lua
    ???
lupa/_lupa.pyx:1328: in lupa._lupa.execute_lua_call
    ???
lupa/_lupa.pyx:250: in lupa._lupa.LuaRuntime.reraise_on_exception
    ???
lupa/_lupa.pyx:1571: in lupa._lupa.py_object_getindex_with_gil
    ???

   ???
E   IndexError: list index out of range
```

The code is working just fine against a real Redis.
Look like it happens due to `ARGV` being `POBJECT` (a Lupa's type?) instead of a regular Lua table.

The same happens when trying to iterate over the results of `redis.call('HGETALL', key)`

bmerry:
@NimrodParasol I'm assuming that this is fixed by #208 and #209. If not please re-open.

PR #209, opened 2018-08-20
Redis operation that returns dict now converted to Lua table when called inside eval operation

NimrodParasol:
During eval operation, results of Redis operations returning dictionaries are now passed as Lua tables instead of python lists
This allows for iteration over them within the Lua script.

This is a fix to the error described in issue [204](https://github.com/jamesls/fakeredis/issues/204)

### DIFF
--- a/test_fakeredis.py
+++ b/test_fakeredis.py
@@ -3346,7 +3346,7 @@ class TestFakeStrictRedis(unittest.TestCase):
         end
         return result
         """
-        val = self.redis.eval(lua, 0, 'foo')
+        val = self.redis.eval(lua, 1, 'foo')
         sorted_val = sorted([val[:2], val[2:]])
         self.assertEqual(
             sorted_val,
--- a/test_fakeredis.py
+++ b/test_fakeredis.py
@@ -3346,7 +3346,7 @@ class TestFakeStrictRedis(unittest.TestCase):
         end
         return result
         """
-        val = self.redis.eval(lua, 0)
+        val = self.redis.eval(lua, 0, 'foo')
         sorted_val = sorted([val[:2], val[2:]])
         self.assertEqual(
             sorted_val,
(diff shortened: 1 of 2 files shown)


nimrod-otonomo:
@bmerry I added the key as you asked. but for future reference please notice that as far as I've seen the tests never use the keys that are passed to them but using hard-coded values within the Lua scripts.

NimrodParasol:
@bmerry  I added the key as you asked. but for future reference please notice that as far as I've seen the tests never use the keys that are passed to them but using hard-coded values within the Lua scripts.

test_fakeredis.py
```diff
@@ -3337,6 +3337,22 @@ def test_eval_hgetall(self):
...
+    def test_eval_hgetall_iterate(self):
+        self.redis.hset('foo', 'k1', 'bar')
+        self.redis.hset('foo', 'k2', 'baz')
+        lua = """
+        local result = redis.call("hgetall", "foo")
+        for i, v in ipairs(result) do
+        end
+        return result
+        """
+        val = self.redis.eval(lua, 0, 'foo')
```
bmerry: Should be 1, not 0 (number of keys passed).

bmerry:
> for future reference please notice that as far as I've seen the tests never use the keys that are passed to them but using hard-coded values within the Lua scripts.

I agree, that's not ideal, but it's legal. According to the redis docs, "All Redis commands must be analyzed before execution to determine which keys the command will operate on. In order for this to be true for EVAL, keys must be passed explicitly." It doesn't matter to redis whether the Lua code uses KEYS or hard-codes them, as long as the redis server is informed about all keys that might be accessed.

test_fakeredis.py
nimrod-otonomo: You are totally right.
Fixed it.

NimrodParasol: You are totally right.
Fixed it.

PR #212, opened 2018-08-27
Optimize _scan()

thedrow:
In case that a regular expression is specified, it is not needed to copy all the keys if some of them do not match the regular expression.
If a regular expression is not provided, there is no need to iterate over the slice. Instead, we simply copy the entire slice.
This is useful in cases where we're dealing with a large number of keys.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -13,7 +13,7 @@ import time
 import types
 import re
 import functools
-from itertools import count
+from itertools import count, islice
 
 import redis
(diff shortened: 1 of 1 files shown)


thedrow:
I'm genuinely surprised there are no "off by one" errors 🤣

thedrow:
Feel free to check. I don't think there will be much of a difference.
I think we gained most of the performance boost we could here.

PR #214, opened 2018-09-05
added pubsub.run_in_thread as it is implemented in redis-py

### LINKED ISSUE #213: `run_in_thread` method is not supported for `fakepubsub` object
My application uses the `run_in_thread` method of the `PubSub` object in `py-redis` to execute a callback function in a separate thread on publish events. This does not appear to be supported in fakeredis currently so I can't write unittests for my application by simply dropping `FakeStrictRedis` in place of `StrictRedis` client.

This module should support the background thread implementation as the real module does.

See the [PubSub Section of the readme](https://github.com/andymccurdy/redis-py#publish--subscribe)

AlexEshoo:
Closes #213

### DIFF
--- a/test_fakeredis.py
+++ b/test_fakeredis.py
@@ -3082,17 +3082,13 @@ class TestFakeStrictRedis(unittest.TestCase):
     def test_pubsub_run_in_thread(self):
         q = Queue()
 
-        def _queue_msg(msg, q):
-            q.put(msg)
-
         pubsub = self.redis.pubsub()
-        pubsub.subscribe(**{"channel": lambda e: _queue_msg(e, q)})
+        pubsub.subscribe(channel=q.put)
         pubsub_thread = pubsub.run_in_thread()
 
         msg = b"Hello World"
         self.redis.publish("channel", msg)
 
-        sleep(1)
         retrieved = q.get()
         self.assertEqual(retrieved["data"], msg)
 
@@ -3100,11 +3096,13 @@ class TestFakeStrictRedis(unittest.TestCase):
         pubsub_thread.join()
         self.assertTrue(not pubsub_thread.is_alive())
 
-        pubsub.subscribe(**{"channel": None})
+        pubsub.subscribe(channel=None)
         with self.assertRaises(redis.exceptions.PubSubError):
             pubsub_thread = pubsub.run_in_thread()
 
-        pubsub.psubscribe(**{"channel": None})
+        pubsub.unsubscribe("channel")
+
+        pubsub.psubscribe(channel=None)
         with self.assertRaises(redis.exceptions.PubSubError):
             pubsub_thread = pubsub.run_in_thread()
 
--- a/test_fakeredis.py
+++ b/test_fakeredis.py
@@ -3078,6 +3078,36 @@ class TestFakeStrictRedis(unittest.TestCase):
         received = q.get()
         self.assertEqual(received['data'], msg)
 
+    @attr('slow')
+    def test_pubsub_run_in_thread(self):
+        q = Queue()
+
+        def _queue_msg(msg, q):
+            q.put(msg)
+
+        pubsub = self.redis.pubsub()
+        pubsub.subscribe(**{"channel": lambda e: _queue_msg(e, q)})
+        pubsub_thread = pubsub.run_in_thread()
+
+        msg = b"Hello World"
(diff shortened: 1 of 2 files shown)


bmerry:
Sorry for the delay on this. I haven't used run_in_thread myself, so I need to get around to reading up on what it does. I'm concerned that this might not be thread-safe given that fakeredis in general is not thread-safe at all.

test_fakeredis.py
```diff
@@ -3078,6 +3078,36 @@ def _listen(pubsub, q):
...
 
+    @attr('slow')
+    def test_pubsub_run_in_thread(self):
+        q = Queue()
+
+        def _queue_msg(msg, q):
+            q.put(msg)
+
+        pubsub = self.redis.pubsub()
+        pubsub.subscribe(**{"channel": lambda e: _queue_msg(e, q)})
```
bmerry: Instead of the lambda and `_queue_msg`, you could just pass `q.put` as the callback.

test_fakeredis.py
```diff
@@ -3078,6 +3078,36 @@ def _listen(pubsub, q):
...
+        pubsub_thread = pubsub.run_in_thread()
+
+        msg = b"Hello World"
+        self.redis.publish("channel", msg)
+
+        sleep(1)
```
bmerry: Is the sleep necessary? q.get should block until the message arrives.

test_fakeredis.py
```diff
@@ -3078,6 +3078,36 @@ def _listen(pubsub, q):
...
+        pubsub_thread.join()
+        self.assertTrue(not pubsub_thread.is_alive())
+
+        pubsub.subscribe(**{"channel": None})
+        with self.assertRaises(redis.exceptions.PubSubError):
+            pubsub_thread = pubsub.run_in_thread()
+
+        pubsub.psubscribe(**{"channel": None})
+        with self.assertRaises(redis.exceptions.PubSubError):
+            pubsub_thread = pubsub.run_in_thread()
```
bmerry: Not sure this test will be effective, since you'll still have the normal `subscribe` to the channel so it will be bailing out due to that rather than the `psubscribe` without handler.

test_fakeredis.py
```diff
@@ -3078,6 +3078,36 @@ def _listen(pubsub, q):
...
+
+        pubsub_thread.stop()
+        pubsub_thread.join()
+        self.assertTrue(not pubsub_thread.is_alive())
+
+        pubsub.subscribe(**{"channel": None})
+        with self.assertRaises(redis.exceptions.PubSubError):
+            pubsub_thread = pubsub.run_in_thread()
+
+        pubsub.psubscribe(**{"channel": None})
```
bmerry: Rather than constructing a dict then `**`ing it, just pass `channel=None`.

bmerry:
I've just noticed that redis-py says "It is not safe to pass PubSub or Pipeline objects between threads." So I'm no longer worried about thread safety, and once my other comments are addressed I'll be happy to merge this.

It also looks like FakePubSubWorkerThread is identical to the redis PubSubWorkerThread (by eye - I didn't diff them). In that case, we could save some code by just importing PubSubWorkerThread.

AlexEshoo:
I agree passing the pubsub objects between threads is not safe, it should be up to the user to decide if what they are doing is safe to do, but in many cases subscriptions can be responded to in a separate thread without needing to modify the pubsub object.

And yes it is a carbon copy other than the name, so importing from redis-py should be fine.

test_fakeredis.py
AlexEshoo: I'm not sure how to pass the event object to the callback if I just use q.put as the callback in this context. I could remove the`_queue_msg` function and simply make it `lambda e: q.put(e)`, however.

AlexEshoo: fixed

test_fakeredis.py
AlexEshoo: fixed

test_fakeredis.py
AlexEshoo: unsubscribed first.

test_fakeredis.py
AlexEshoo: fixed

bmerry:
Thanks for the updates. It all looks good, but just needs a fix for a merge conflict with some thread-safety improvements I merged this morning.

PR #216, opened 2018-09-17
Update behavior of lock to behave closer to redis lock

taylor-cedar:
I updated a project to use fakeredis and noticed some tests failing because the fakeredis lock implementation doesn't actually behave like the redis lock. If you create 2 locks with the same key, it doesn't actually fail to acquire the lock.

I added more tests and updated the implementation to handle multiple locks on the same key.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -341,7 +341,7 @@ class _Lock(object):
             raise LockError("Cannot release an unlocked lock")
 
         if _decode(self.redis.get(self.name)) != self.id:
-            raise LockError("Cannot extend a lock that's no longer owned")
+            raise LockError("Cannot release a lock that's no longer owned")
 
         self.redis.delete(self.name)
         self.id = None
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -12,6 +12,7 @@ import time
 import types
 import re
 import functools
+import threading
 from itertools import count, islice
 from uuid import uuid4
 
@@ -311,6 +312,9 @@ def _compile_pattern(pattern):
     return re.compile(regex, re.S)
 
 
+threading_lock = threading.Lock()
+
+
 class _Lock(object):
     def __init__(self, redis, name, timeout):
         self.redis = redis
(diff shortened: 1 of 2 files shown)


fakeredis.py
```diff
@@ -326,11 +326,18 @@ def __exit__(self, exc_type, exc_value, traceback):
         self.release()
 
     def acquire(self, blocking=True, blocking_timeout=None):
-        return self.lock.acquire(blocking)
+        owner_id = _decode(self.redis.get(self.name))
+        can_acquire = not owner_id or owner_id == self.id
+        if can_acquire:
+            self.redis.set(self.name, self.id, ex=self.timeout)
+        elif blocking:
+            raise ValueError('fakeredis can\'t do blocking locks')
```
taylor-cedar: I originally had an implementation to sleep and wait for the lock to be released, but eventually changed it to an error because blocking would likely wait forever or until the `blocking_timeout`. In either case I think it's an error and it's better to expose an exception instead of an infinite wait.

taylor-cedar:
@bmerry Yes, this definitely isn't thread safe (as you said), but it wasn't before anyways. Can you merge it in for now and make it a minor version release for now? This isn't making anything worse and it fixes several issues that currently exist.

Thread safety is something that is important, but it seems outside the scope of this PR.

bmerry:
The old code used threading.Lock, so it looks like it was thread-safe, as long as you used the same Lock object across threads.

taylor-cedar:
@bmerry Yes the original code was thread safe because it didn't really do any locking at all. It was basically a noop. I guess this could cause random issues if used in a threading environment. I added code to make it thread safe

bmerry:
Thanks, I'll try to find some time this week to review this.

bmerry:
I've started a branch called "thread-safe" where I'm working on thread safety, and I've also replaced `_Lock` with a copy of redis.lock.Lock plus a few bug fixes. Give it a try and let me know if it works for your use case. I may also borrow some tests from this PR still.

bmerry:
I'm closing this because I've put in an implementation that's a copy of redis-py's plus bug fixes, and also taken some of your unit tests (thanks). Let me know if there are still issues.

PR #222, opened 2018-11-06
Python 3.7

owenstranathan:
It works out of the box. Only thing is deprecation warnings that are annoying in tests.

Given this: 
https://docs.python.org/3/whatsnew/3.7.html#id3

Getting ABCs from `collections` is depricated in 3.8 so this preemptively fixes that and also get rid of those pesky aforementioned deprecation warnings.

Also adds a Pipfile for those of us who have adopted Pipenv.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,15 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+
+[dev-packages]
+tox = "*"
+nose = "==1.3.4"
+
+[packages]
+"flake8" = "<3.0.0"
+redis = "==2.10.6"
+lupa = "==1.6"
+future = "==0.16.0"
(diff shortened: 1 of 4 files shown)


owenstranathan:
@bmerry Added Explicit `except ImportError`. Also I would be happy to maintain the Pipfile.

bmerry:
Great. Travis failed for some reason, but it doesn't seem like it's related to your change at all so I'm rerunning it. Assuming that passes I'll merge.

@jamesls, if you're happy could you give @owenstranathan write access? It seems I don't have permission.

bmerry:
I suspect the Python 3.7 problem is because Travis has switched to redis 5.0 (but presumably only on xenial, not trusty). Master is also failing. So I'm going to merge anyway.

bmerry:
@owenstranathan I've got a branch called `rewrite` which has new and updated dependencies. Could you contribute an updated Pipfile for it?

Issue #223, opened 2018-11-08
__delitem__ missing


viljoviitanen:
Noticed fakeredis does not support "del" operator, i.e. it does not implement `__delitem__` , while real redis does.

So code like this works on real redis:
```python
import redis
r = redis.StrictRedis(host='localhost', port=6379, db=0)
r['foo'] = 'bar'
print (r['foo'])  #bar
del r['foo']
print (r['foo'])  #KeyError: 'foo'
```
but fails on fakeredis:
```python
import fakeredis
r = fakeredis.FakeStrictRedis()
r['foo'] = 'bar'
print (r['foo'])  #bar
del r['foo']      #AttributeError: __delitem__
#r.delete('foo')  #this would work
```

My guess on this is that even if  `__delitem__` is there in _StrKeyDict, because of name mangling, this happens.

If there's interest in getting this fixed, I can probably do a clean enough PR.

bmerry:
> If there's interest in getting this fixed, I can probably do a clean enough PR.

Yes, PRs are very welcome. I'm guessing it should be a pretty thin wrapper around `.delete`.

PR #224, opened 2018-11-08
Add __delitem__

viljoviitanen:
Issue #223 
A separate travis-ci run that shows the test first failing without implementation:
https://travis-ci.org/viljoviitanen/fakeredis/builds/452512706

Also to make things sure: I allow my small contribution to be distributed under any licence and assign copyright to current copyright owners.

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -980,6 +980,9 @@ class FakeStrictRedis(object):
                 continue
         return deleted
 
+    def __delitem__(self, name):
+        return self.delete(name)
+
     @_locked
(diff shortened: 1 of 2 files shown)


fakeredis.py
```diff
@@ -980,6 +980,10 @@ def delete(self, *names):
                 continue
         return deleted
 
+    @_locked
```
bmerry: `self.delete` already takes the lock, so it's unnecessary for this function to be locked.

viljoviitanen:
Force pushed a commit without the unnecessary @_locked decorator.

bmerry:
Thanks, merged!

Issue #119, opened 2016-06-21
FakeStrictRedis ttl should not return None


marakal:
StrictRedis' ttl and pttl methods return -1 when expire is not set and -2 when key does not exist ([documentation of python redis](https://pypi.python.org/pypi/redis), [documentation of redis](http://redis.io/commands/ttl)), but FakeStrictRedis' methods return None instead.

```
In [1]: r = redis.StrictRedis(db=0)

In [2]: r.ttl("foo")
Out[2]: -2

In [3]: f = fakeredis.FakeStrictRedis()

In [4]: f.ttl("foo")
Out[4]: None
```

EricCline:
+1

gazoon:
+1

jamesls:
Fixed via https://github.com/jamesls/fakeredis/pull/134

PR #228, opened 2018-12-17
ttl should not return None.

amw:
`pttl(key) / 1000` breaks my tests when this returns None.

Similar to #119

I tried adding a test, but in a single threaded env the test passes with or without my patch.

```diff
diff --git a/test_fakeredis.py b/test_fakeredis.py
index eea5f63..9c7eabb 100644
--- a/test_fakeredis.py
+++ b/test_fakeredis.py
@@ -3334,6 +3334,13 @@ class TestFakeStrictRedis(unittest.TestCase):
         self.assertEqual(self.redis.get('foo'), None)
         self.assertEqual(self.redis.ttl('foo'), -2)
 
+    def test_ttl_should_return_minus_two_for_expired_key(self):
+        self.redis.set('foo', 'bar')
+        self.assertEqual(self.redis.get('foo'), b'bar')
+        self.redis.pexpire('foo', 1)
+        sleep(0.002)
+        self.assertEqual(self.redis.ttl('foo'), -2)
+
     def test_pttl_should_return_minus_one_for_non_expiring_key(self):
         self.redis.set('foo', 'bar')
         self.assertEqual(self.redis.get('foo'), b'bar')
@@ -3343,6 +3350,13 @@ class TestFakeStrictRedis(unittest.TestCase):
         self.assertEqual(self.redis.get('foo'), None)
         self.assertEqual(self.redis.pttl('foo'), -2)
 
+    def test_pttl_should_return_minus_two_for_expired_key(self):
+        self.redis.set('foo', 'bar')
+        self.assertEqual(self.redis.get('foo'), b'bar')
+        self.redis.pexpire('foo', 1)
+        sleep(0.002)
+        self.assertEqual(self.redis.pttl('foo'), -2)
+
     def test_persist(self):
         self.redis.set('foo', 'bar', ex=20)
         self.redis.persist('foo')
```

### DIFF
--- a/fakeredis.py
+++ b/fakeredis.py
@@ -940,7 +940,7 @@ class FakeStrictRedis(object):
 
         now = datetime.now()
         if now > exp_time:
-            return None
+            return -2  # expired key acts as if it didn't exist
         else:
             return long(round(((exp_time - now).days * 3600 * 24 +
                         (exp_time - now).seconds +


bmerry:
I've been working on a rewrite of fakeredis, in branch `rewrite`. Could you check if the issue still exists there?

amw:
It's difficult to test. Is access to keys protected with a lock like in real Redis or can a different thread execute a second command in parallel? If two threads can work with a key at the same time then you could have a problem. Here's your current code:

```python
    def _ttl(self, key, scale):
        if not key:
            return -2
        elif key.expireat is None:  # here we access the property for the first time
            return -1
        else:
            # here we access it a second time, but different thread could change it to `None` in between
            return int(round((key.expireat - self._db.time) * scale))
```

A safer version would be:

```python
    def _ttl(self, key, scale):
        if not key:
            return -2

        expireat = key.expireat

        if expireat is None:
            return -1

        return int(round((expireat - self._db.time) * scale))
```

bmerry:
That's actually something else I fixed in the rewrite: it now samples the clock at the start of executing the command, and then uses that cached value throughout executing the command (and a lock is held for the entire time, so another command cannot update the time). So I believe it should work even without your change.

PR #229, opened 2018-12-18
Protect _pubsubs list from concurrent changes.

amw:
My tests were failing when I was using fake pubsub in a multithreaded app. This patch protects all changes to `_pubsubs` list with a lock.

I'm not sure how to go about tests for these multithreading issues.


bmerry:
Thanks for the contribution. However, fakeredis has gone through a rewrite recently (see `rewrite` branch) and this patch will no longer apply. I think this should no longer be an issue (since all commands now take a lock), but let me know if your tests still have issues with 1.0rc1.

PR #231, opened 2019-01-12
Fix setup.py, fakeredis is directory/package now

onovy:


bmerry:
Thanks! Apart from this, is 1.0b1 working for you?

PR #236, opened 2019-03-13
Support for redis-py 3.2 

### LINKED ISSUE #235: Does not work with redis-py 3.2
If I install 1.0.1 and try to execute what doc says I obtain the following error:
```Python
In [1]: import fakeredis                                                                                                                                                                                                                                                                                                   

In [2]: r = fakeredis.FakeStrictRedis()                                                                                                                                                                                                                                                                                    

In [3]: r.set('a', 'b')                                                                                                                                                                                                                                                                                                    
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-9aae901f3e7b> in <module>
----> 1 r.set('a', 'b')

~/.local/share/virtualenvs/api-fKmI9ai6/lib/python3.6/site-packages/redis/client.py in set(self, name, value, ex, px, nx, xx)
   1449         if xx:
   1450             pieces.append('XX')
-> 1451         return self.execute_command('SET', *pieces)
   1452 
   1453     def __setitem__(self, name, value):

~/.local/share/virtualenvs/api-fKmI9ai6/lib/python3.6/site-packages/elasticapm/instrumentation/packages/base.py in call_if_sampling(self, module, method, wrapped, instance, args, kwargs)
    103         transaction = get_transaction()
    104         if not transaction or not transaction.is_sampled:
--> 105             return wrapped(*args, **kwargs)
    106         else:
    107             return self.call(module, method, wrapped, instance, args, kwargs)

~/.local/share/virtualenvs/api-fKmI9ai6/lib/python3.6/site-packages/elasticapm/instrumentation/packages/base.py in call_if_sampling(self, module, method, wrapped, instance, args, kwargs)
    103         transaction = get_transaction()
    104         if not transaction or not transaction.is_sampled:
--> 105             return wrapped(*args, **kwargs)
    106         else:
    107             return self.call(module, method, wrapped, instance, args, kwargs)

~/.local/share/virtualenvs/api-fKmI9ai6/lib/python3.6/site-packages/redis/client.py in execute_command(self, *args, **options)
    770         pool = self.connection_pool
    771         command_name = args[0]
--> 772         connection = pool.get_connection(command_name, **options)
    773         try:
    774             connection.send_command(*args)

~/.local/share/virtualenvs/api-fKmI9ai6/lib/python3.6/site-packages/redis/connection.py in get_connection(self, command_name, *keys, **options)
    992         try:
    993             # ensure this connection is connected to Redis
--> 994             connection.connect()
    995             # connections that the pool provides should be ready to send
    996             # a command. if not, the connection was either returned to the

~/.local/share/virtualenvs/api-fKmI9ai6/lib/python3.6/site-packages/redis/connection.py in connect(self)
    498 
    499         self._sock = sock
--> 500         self._selector = DefaultSelector(sock)
    501         try:
    502             self.on_connect()

~/.local/share/virtualenvs/api-fKmI9ai6/lib/python3.6/site-packages/redis/selector.py in DefaultSelector(sock)
    194         else:
    195             raise RedisError('Platform does not support any selectors')
--> 196     return _DEFAULT_SELECTOR(sock)

~/.local/share/virtualenvs/api-fKmI9ai6/lib/python3.6/site-packages/redis/selector.py in __init__(self, sock)
    122             super(PollSelector, self).__init__(sock)
    123             self.read_poller = select.poll()
--> 124             self.read_poller.register(sock, self._READ_POLLER_MASK)
    125             self.ready_poller = select.poll()
    126             self.ready_poller.register(sock, self._READY_POLLER_MASK)

TypeError: argument must be an int, or have a fileno() method.
```
Any idea of what is wrong?

Raab70:
Fixes #235 

redis-py 3.2 introduces the new concept of a selector. Herein we introduce a `FakeSelector` which marks our fake connections as always ready. The `FakeSocket` also needs to return an integer from `fileno()` since this will be checked when importing `redis.selector`. 

Let me know if you have any feedback, I've run the full test suite with redis-py 2.10.6, 3.1.0 and 3.2.0. I did not contribute a new test since existing tests fail with 3.2.0.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -371,6 +371,10 @@ they have all been tagged as 'slow' so you can skip them by running::
 Revision history
 ================
 
+1.0.3
+-----
+- `#235 <https://github.com/jamesls/fakeredis/issues/235>`_ Support for ``redis==3.2``
+
 1.0.2
 -----
 - `#235 <https://github.com/jamesls/fakeredis/issues/235>`_ Depend on ``redis<3.2``
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -646,6 +646,9 @@ class FakeSocket(object):
         self._parser.close()
 
     def fileno(self):
+        # Our fake socket must return an integer from `FakeSocket.fileno()` since a real selector
+        # will be created. The value does not matter since we replace the selector with our own
+        # `FakeSelector` before it is ever used.
         return 0
 
     def close(self):
@@ -2405,16 +2408,15 @@ class _DummyParser(object):
 
 
 # Redis <3.2 will not have a selector
-if hasattr(redis, 'selector'):
-    base_selector = redis.selector.BaseSelector
-else:
-    class FakeBaseSelector:
+try:
+    from redis.selector import BaseSelector
+except ImportError:
+    class BaseSelector(object):
         def __init__(self, sock):
             pass
-    base_selector = FakeBaseSelector
 
 
-class FakeSelector(base_selector):
+class FakeSelector(BaseSelector):
     def check_can_read(self, timeout):
         return True
 
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -645,6 +645,9 @@ class FakeSocket(object):
     def shutdown(self, flags):
         self._parser.close()
 
+    def fileno(self):
+        return 0
+
     def close(self):
         with self._server.lock:
             for subs in self._server.subscribers.values():
@@ -2401,6 +2404,24 @@ class _DummyParser(object):
         pass
 
 
+# Redis <3.2 will not have a selector
+if hasattr(redis, 'selector'):
+    base_selector = redis.selector.BaseSelector
+else:
+    class FakeBaseSelector:
+        def __init__(self, sock):
+            pass
+    base_selector = FakeBaseSelector
+
+
+class FakeSelector(base_selector):
+    def check_can_read(self, timeout):
+        return True
+
+    def check_is_ready_for_command(self, timeout):
+        return True
+
+
 class FakeConnection(redis.Connection):
     description_format = "FakeConnection<db=%(db)s>"
 
@@ -2421,6 +2442,11 @@ class FakeConnection(redis.Connection):
         # override them.
         self._parser = _DummyParser()
         self._sock = None
+
+    def connect(self):
+        super().connect()
+        # The selector is set in redis.Connection.connect() after _connect() is called
+        self._selector = FakeSelector(self._sock)
 
     def _connect(self):
         if not self._server.connected:
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -2444,7 +2444,7 @@ class FakeConnection(redis.Connection):
         self._sock = None
 
     def connect(self):
-        super().connect()
+        super(FakeConnection, self).connect()
         # The selector is set in redis.Connection.connect() after _connect() is called
         self._selector = FakeSelector(self._sock)
 
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -2413,12 +2413,26 @@ try:
 except ImportError:
     class BaseSelector(object):
         def __init__(self, sock):
-            pass
+            self.sock = sock
 
 
 class FakeSelector(BaseSelector):
     def check_can_read(self, timeout):
-        return True
+        if self.sock.responses.qsize():
+            return True
+        if timeout <= 0:
+            return False
+
+        # A sleep/poll loop is easier to mock out than messing with condition
+        # variables.
+        start = time.time()
+        while True:
+            if self.sock.responses.qsize():
+                return True
+            time.sleep(0.01)
+            now = time.time()
+            if now > start + timeout:
+                return False
 
     def check_is_ready_for_command(self, timeout):
         return True
@@ -2462,7 +2476,7 @@ class FakeConnection(redis.Connection):
             self.connect()
         if self._sock.responses.qsize():
             return True
-        while timeout <= 0:
+        if timeout <= 0:
             return False
 
         # A sleep/poll loop is easier to mock out than messing with condition
@@ -2471,6 +2485,7 @@ class FakeConnection(redis.Connection):
         while True:
             if self._sock.responses.qsize():
                 return True
+            time.sleep(0.01)
             now = time.time()
             if now > start + timeout:
                 return False
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ attrs==18.2.0
 enum34==1.1.6; python_version<"3"
 nose==1.3.7
 hypothesis==3.86.4
-redis==2.10.6
+redis==3.2.0
 lupa==1.7
 future==0.17.1
 sortedcontainers==2.1.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ attrs==18.2.0
 enum34==1.1.6; python_version<"3"
 nose==1.3.7
 hypothesis==3.86.4
-redis==3.2.0
+redis==3.2.1
 lupa==1.7
 future==0.17.1
 sortedcontainers==2.1.0
--- a/test_fakeredis.py
+++ b/test_fakeredis.py
@@ -3286,6 +3286,7 @@ class TestFakeStrictRedis(unittest.TestCase):
 
         pubsub_thread.stop()
         # Newer versions of redis wait for an unsubscribe message, which sometimes comes early
+        # https://github.com/andymccurdy/redis-py/issues/1150
         if pubsub.channels:
             pubsub.channels = {}
         pubsub_thread.join()
--- a/test_fakeredis.py
+++ b/test_fakeredis.py
@@ -3285,6 +3285,9 @@ class TestFakeStrictRedis(unittest.TestCase):
         self.assertEqual(retrieved["data"], msg)
 
         pubsub_thread.stop()
+        # Newer versions of redis wait for an unsubscribe message, which sometimes comes early
+        if pubsub.channels:
+            pubsub.channels = {}
         pubsub_thread.join()
         self.assertTrue(not pubsub_thread.is_alive())
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ python:
 env:
   - REDIS_PY=2.10.6
   - REDIS_PY=3.0.1
-  - REDIS_PY=3.2.0
+  - REDIS_PY=3.1.0
 matrix:
   include:
     - python: 3.7
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ python:
 env:
   - REDIS_PY=2.10.6
   - REDIS_PY=3.0.1
+  - REDIS_PY=3.2.0
 matrix:
   include:
     - python: 3.7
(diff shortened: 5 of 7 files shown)


bmerry:
Thanks! I'll take a look sometime soon hopefully (I don't get much time for fakeredis). This looks a lot simpler than I was afraid it was going to be. I thought I might have to implement real sockets and stuff data into them in redis protocol format.

Raab70:
@bmerry Thanks, I see that the Travis builds are timing out on redis-py 3.2.0, which I added, which tells me I'm still not patching something correctly. I'll continue digging. Tests run locally with 3.2 when I turn off `slow` tests but if I run all of them I get: 

```
Falsifying example: run_state_machine(factory=Machine, data=data(...))
Step #1: {'fields': [b'', b'\x00'],
 'keys': [b'', b'\x00'],
 'scores': [1.0, 0.0],
 'values': [b'', b'\x00']}
Step #2: []
Step #3: Command('zadd', b'', 1.0, b'')
Step #4: Command('zunionstore', b'', 7, b'', b'', b'', b'', b'', b'', b'', 'weights', b'0.0', b'1.0', b'1.0', b'9007199254740992.0', b'0.0', b'0.0', b'0.0')
Step #5: Command('zrange', b'', 0, 0, 'withscores')
...
AssertionError: Lists differ: [b'', b'9007199254740994'] != [b'', b'9007199254740992']
```

I've never used hypothesis although I know the premise. Just seems odd that the small amount of connection changes would cause this unless there is still some code that is actually reading data from sockets somehow...

bmerry:
hypothesis is randomised, and occasionally it throws up new errors after apparently working for a long time (and it then remembers the failing test case and keeps throwing it up). In this case my guess is that it comes down to the order in which the arithmetic is done not being the same as redis (in double precision, 1.0+1.0+9007199254740992.0 != 9007199254740992.0+1.0+1.0 due to rounding errors). Don't worry about it for now.

bmerry:
> Don't worry about it for now.

By which I mean don't worry about that specific hypothesis failure. Tests hanging is more of a concern.

Raab70:
@bmerry Thanks I was hoping it was something like that. 

As for the stalled build I can reproduce on a Docker image of trusty so I'll try to see where it's hung.

Raab70:
I think I'm getting close but I'm not immediately seeing how to fix so I'll document. I can reproduce with:

`docker run -v $(pwd):/usr/src/app ubuntu:trusty bash`
`apt-get update && apt-get install python3-pip redis-server`
`service redis-server start`
`python -m pip install --upgrade pip`
`pip install virtualenv`
`virtualenv venv`
`source venv/bin/activate`
`pip install -r requirements.txt`
`pip install coverage redis==3.1.0`
`nosetests -v test_fakeredis:TestRealStrictRedis.test_pubsub_run_in_thread` 
That last command will hang but then if you CTRL-C it will print that the test passed and the stacktrace of trying to join on the thread. It does not happen every time but I would say most times it does. 

I'm thinking the issue is around [here](https://github.com/andymccurdy/redis-py/compare/3.0.1...3.1.0#diff-085cbe85ef9bc2ad832d79163eb39172R3117) but that still doesn't explain why I can run this on OSX without issue.

bmerry:
Have you tried it with fakeredis master? I've not tried with 3.1.0 myself, so it might not be anything you introduced.

bmerry:
Hmm, I guess I ought to get the tests working with 3.1 first, before we try to support 3.2. Unfortunately that's going to delay this PR a bit.

Raab70:
@bmerry That is unfortunate, can we not merge this and make a release, maybe an `rc` with some documentation around the pubsub limitations. Then we can create a new issue for tracking the pubsub incompatibilities with redis-py>=3.1. I'll keep digging on the pubsub threading but it's extremely difficult to trace given that it isn't reliably reproducible and differs between OS.

Raab70:
I have narrowed this down to this PR: https://github.com/andymccurdy/redis-py/pull/1105

This is really just a race condition in the test, `pubsub_thread.stop()` now runs the unsubscribe command against redis but does not update `self.channels` until it receives a message. So what happens is that the pubsub thread can receive the response from redis before the channel is added to `self.pending_unsubscribe_channels` and the channel is then not unsubscribed. It's interesting that I have been unable to reproduce this with a single script but only from within `nosetests`. Here is some output of some print statements I added to the redis client [near here](https://github.com/andymccurdy/redis-py/blob/3.0.1/redis/client.py#L3098)

```
Handling message:  {'data': 1, 'type': 'subscribe', 'pattern': None, 'channel': b'channel'}
Handling message:  {'data': b'Hello World', 'type': 'message', 'pattern': None, 'channel': b'channel'}
Handling message:  {'data': 0, 'type': 'unsubscribe', 'pattern': None, 'channel': b'channel'}
Message is unsubscribe type
Checking if channel:  b'channel' is in:  set()
```
You can see that the unsubscribed channels set is empty, so the channel never gets removed from `pubsub.channels` and the thread keeps listening. The simple fix is probably to explicitly set `pubsub.channels = {}` but let me know what you think. I could open a ticket but again I can only really reproduce in `nosetests`, I have a script which does the same thing but it happens less frequently although I have gotten it to hang. Not sure there's much they could do about it anyways.

bmerry:
Good work - thanks for investigating. I'll have to look at the code before this makes sense to me. Hopefully I'll have time next week to review things.

fakeredis/_server.py
```diff
@@ -2401,6 +2404,24 @@ def on_connect(self, connection):
...
+else:
+    class FakeBaseSelector:
+        def __init__(self, sock):
+            pass
+    base_selector = FakeBaseSelector
+
+
+class FakeSelector(base_selector):
+    def check_can_read(self, timeout):
+        return True
```
bmerry: That doesn't look right. It should do something similar to `FakeSocket.can_read` (obviously sharing the common code).

fakeredis/_server.py
```diff
@@ -2422,6 +2443,11 @@ def __init__(self, server, db=0, password=None,
         self._parser = _DummyParser()
         self._sock = None
 
+    def connect(self):
+        super(FakeConnection, self).connect()
+        # The selector is set in redis.Connection.connect() after _connect() is called
+        self._selector = FakeSelector(self._sock)
```
bmerry: I'm not totally comfortable with letting the base class build a DefaultSelector with a fileno that might not even be valid, and is most likely not a socket. It works for now on Linux, but I'm not sure how portable it is, and if redis changes the default selector implementation (e.g. to use epoll) then I imagine it could break. I'm not sure what the best fix is though - possibly monkey-patching DefaultSelector so that it will return a FakeSelector when given a FakeSocket, but monkey-patching also tends to be quite fragile.

Any thoughts?

requirements.txt
```diff
@@ -7,7 +7,7 @@ attrs==18.2.0
 enum34==1.1.6; python_version<"3"
 nose==1.3.7
 hypothesis==3.86.4
-redis==2.10.6
+redis==3.2.0
```
bmerry: Might as well go to 3.2.1.

test_fakeredis.py
```diff
@@ -3285,6 +3285,9 @@ def test_pubsub_run_in_thread(self):
         self.assertEqual(retrieved["data"], msg)
 
         pubsub_thread.stop()
+        # Newer versions of redis wait for an unsubscribe message, which sometimes comes early
```
bmerry: Can you add a link to the redis issue you filed.

fakeredis/_server.py
```diff
@@ -645,6 +645,9 @@ def __init__(self, server):
     def shutdown(self, flags):
         self._parser.close()
 
+    def fileno(self):
+        return 0
```
bmerry: Can you add a comment about why this is needed and why the return value doesn't matter too much.

fakeredis/_server.py
```diff
@@ -2401,6 +2404,24 @@ def on_connect(self, connection):
...
 
 
+# Redis <3.2 will not have a selector
+if hasattr(redis, 'selector'):
+    base_selector = redis.selector.BaseSelector
+else:
+    class FakeBaseSelector:
+        def __init__(self, sock):
+            pass
+    base_selector = FakeBaseSelector
```
bmerry: I would write this bit like this:
```python
try:
    from redis.selector import BaseSelector
except ImportError:
    class BaseSelector(object):
        def __init__(self, sock):
            pass


class FakeSelector(BaseSelector):
```

fakeredis/_server.py
```diff
@@ -2401,6 +2404,24 @@ def on_connect(self, connection):
         pass
 
 
+# Redis <3.2 will not have a selector
+if hasattr(redis, 'selector'):
+    base_selector = redis.selector.BaseSelector
+else:
+    class FakeBaseSelector:
```
bmerry: Inherit from object (otherwise this becomes an old-style class in Python 2).

README.rst
```diff
@@ -371,6 +371,10 @@ they have all been tagged as 'slow' so you can skip them by running::
 Revision history
 ================
 
+1.0.3
+-----
+- `#235 <https://github.com/jamesls/fakeredis/issues/235>`_ Support for ``redis==3.2``
```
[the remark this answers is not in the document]
Raab70: I was following the format of the previous entry which references the issue and not the PR, which is #235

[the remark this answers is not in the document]
bmerry: Oh right, I'm with you now.

fakeredis/_server.py
Raab70: I'm not seeing a `FakeSocket.can_read`, there is a `FakeConnection.can_read` is that what you were referring to? That appears to handle ensuring the socket is connected, the selector handles checking if the socket has data: https://github.com/andymccurdy/redis-py/blob/3.2.0/redis/selector.py#L49 and https://github.com/andymccurdy/redis-py/blob/3.2.0/redis/selector.py#L96

fakeredis/_server.py
Raab70: I agree, this is not ideal. A real selector will be created on our fake socket which either uses `select.select` or `select.poll` The selector module does a good job of ensuring compatibility however and `0` is a valid socket on any system.

fakeredis/_server.py
bmerry: That's the one I meant.

Your `check_can_read` isn't checking if there is data available though, it's just returning true. It should check if `FakeSocket.responses.qsize()` is non-zero to determine whether there is data available, and wait up to the timeout for it to become non-zero.

I see that `FakeConnection.can_read` doesn't actually have a sleep in its `while True` loop; I'd intended to sleep for a short time (maybe 0.01 seconds).

fakeredis/_server.py
bmerry: `0` isn't normally a socket, but it is a file handle (stdin). I suspect that Windows won't be too happy (it doesn't treat sockets and file handles as interchangeably as POSIX does), but I'm not sure if redis-py runs on Windows anyway (neither I nor my colleagues use Windows, so I don't have a way to check).

It's also perfectly possible for a process to close FD 0 (I think some batch jobs do to detach from the terminal), or even to open a new file-like thing in their place. The good news is that I did a test with this and `select.poll` didn't seem to mind. I'd say let's just add some comments and move on, and fix it if it becomes a problem later.

bmerry: Ah, I see you added the comments already.

fakeredis/_server.py
Raab70: ping @bmerry, I pushed a fix for this.

bmerry:
Thanks for the work on this! There are a few minor tweaks I want to make, but I'll just merge and then do them on master.

PR #244, opened 2019-07-29
Rename assertRaisesRegexp to assertRaisesRegex, closes #241

### LINKED ISSUE #241: DeprecationWarning of assertRaisesRegexp in test_fakeredis.py
```
test_fakeredis.py:3621: DeprecationWarning: Please use assertRaisesRegex instead.
  redis.ResponseError, 'value is not an integer or out of range'):
```

isms:
Reference: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex

### DIFF
--- a/test_fakeredis.py
+++ b/test_fakeredis.py
@@ -407,7 +407,7 @@ class TestFakeStrictRedis(unittest.TestCase):
 
     def test_incrbyfloat_bad_type(self):
         self.redis.set('foo', 'bar')
-        with self.assertRaisesRegexp(redis.ResponseError, 'not a valid float'):
+        with self.assertRaisesRegex(redis.ResponseError, 'not a valid float'):
             self.redis.incrbyfloat('foo', 1.0)
         self.redis.rpush('foo2', 1)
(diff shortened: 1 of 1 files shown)


bmerry:
This is failing all the Travis tests.

isms:
@bmerry Oddly they pass locally with Python 3.6. I'm not sure what the issue is.

bmerry:
The old assertRaisesRegexp spelling is the only version supported in Python 2 (and presumably why Python 3 still supports it). Is the deprecation warning causing issues anywhere? Since it's only in the test code I'd prefer to leave it as is rather than adding complexity. Once Python 2 reaches end-of-life I plan to remove Python 2 support and at that time it could be addressed.

isms:
Fair enough, seems like a good plan for backwards compatibility until 2.7
EOL.

adamantike:
@isms can you reopen this PR, and update it by using [`six.assertRaisesRegex`](https://six.readthedocs.io/#six.assertRaisesRegex)?

PR #246, opened 2019-08-04
Migrate project to PyTest

adamantike:
`nose` is a deprecated project since 2015. Migrate to `pytest`, which will allow to add plugins and improve project's test suite.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,15 +1,29 @@
-flake8==3.6.0
+atomicwrites==1.3.0
+attrs==18.2.0
 configparser==3.5.0; python_version<"3"
+contextlib2==0.5.5; python_version<"3"
+coverage==4.5.4
+enum34==1.1.6; python_version<"3"
+flake8==3.6.0
+funcsigs==1.0.2; python_version<"3"
+future==0.17.1
+hypothesis==3.86.4
+importlib-metadata==0.20
+lupa==1.8
 mccabe==0.6.1
+more-itertools==5.0.0
+packaging==19.1
+pathlib2==2.3.4; python_version<"3"
+pluggy==0.12.0
+py==1.8.0
 pycodestyle==2.4.0
 pyflakes==2.0.0
-attrs==18.2.0
-enum34==1.1.6; python_version<"3"
+pyparsing==2.4.2
 pytest==4.6.4
 pytest-cov==2.7.1
-hypothesis==3.86.4
 redis==3.2.1
-lupa==1.8
-future==0.17.1
-sortedcontainers==2.1.0
+scandir==1.10.0; python_version<"3"
 six==1.12.0
+sortedcontainers==2.1.0
+wcwidth==0.1.7
+zipp==0.6.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,8 @@ pycodestyle==2.4.0
 pyflakes==2.0.0
 attrs==18.2.0
 enum34==1.1.6; python_version<"3"
-nose==1.3.7
+pytest==4.6.4
+pytest-cov==2.7.1
 hypothesis==3.86.4
 redis==3.2.1
 lupa==1.8
--- a/test_fakeredis.py
+++ b/test_fakeredis.py
@@ -52,22 +52,8 @@ def redis_must_be_running(cls):
     return cls
 
 
-def redis2_only(func):
-    @wraps(func)
-    def wrapper(*args, **kwargs):
-        if REDIS3:
-            pytest.skip("Test is only applicable to redis-py 2.x")
-        return func(*args, **kwargs)
-    return wrapper
-
-
-def redis3_only(func):
-    @wraps(func)
-    def wrapper(*args, **kwargs):
-        if not REDIS3:
-            pytest.skip("Test is only applicable to redis-py 3.x+")
-        return func(*args, **kwargs)
-    return wrapper
+redis2_only = pytest.mark.skipif(REDIS3, "Test is only applicable to redis-py 2.x")
+redis3_only = pytest.mark.skipif(not REDIS3, "Test is only applicable to redis-py 3.x")
 
 
 def key_val_dict(size=100):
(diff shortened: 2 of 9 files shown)


requirements.txt
```diff
@@ -5,7 +5,8 @@ pycodestyle==2.4.0
 pyflakes==2.0.0
 attrs==18.2.0
 enum34==1.1.6; python_version<"3"
-nose==1.3.7
+pytest==4.6.4
+pytest-cov==2.7.1
```
bmerry: requirements.txt should pin *all* the requirements, including dependencies on pytest. If you create a fresh virtual environment and then `pip install -r requirements.txt` you should be able to determine the dependencies that need pinning.

test_fakeredis.py
```diff
@@ -53,7 +52,7 @@ def redis2_only(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
         if REDIS3:
-            raise SkipTest("Test is only applicable to redis-py 2.x")
+            pytest.skip("Test is only applicable to redis-py 2.x")
         return func(*args, **kwargs)
```
bmerry: This'll work, but I think it should be possible to simplify it a lot using [skipif](https://doc.pytest.org/en/latest/skipping.html#id1) - something like (untested)
```python
redis2_only = pytest.mark.skipif(REDIS3, "Test is only applicable to redis-py 2.x")
```

test_fakeredis.py
```diff
@@ -159,15 +158,15 @@ def test_future_newbytes(self):
         try:
             from builtins import bytes
         except ImportError:
-            raise SkipTest('future.types not available')
+            pytest.skip('future.types not available')
```
bmerry: I think this could be simplified using [pytest.importorskip](https://doc.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency).

test_fakeredis.py
```diff
@@ -4013,7 +4012,7 @@ class TestFakeRedis(unittest.TestCase):
 
     def setUp(self):
         if REDIS3:
-            raise SkipTest('Legacy redis class does not apply to redis-py 3+')
+            pytest.skip('Legacy redis class does not apply to redis-py 3+')
```
bmerry: If `redis3_only` is rewritten as a pytest marker then it can be used to decorate the class.

requirements.txt
adamantike: I still don't get why this project has a `requirements.txt` file, which is intended for deployable projects. To create a fresh virtual environment and install all the dependencies `fakeredis` needs, `pip install -e .[testing]` should be used.

As `fakeredis` is a shared library that must work with different Python versions, and also with relaxed dependency versions, having a `requirements.txt` file goes against that purpose.

Please let me know if there's something I'm missing here. Otherwise, I can remove the `requirements.txt` file, and update the documentation to use the `setup.py` file to install requirements (or even better, Tox).

bmerry: It's used for Travis, to ensure that tests are repeatable (and is also handy to locally test in the same environment that is tested on Travis).

test_fakeredis.py
```diff
@@ -163,19 +149,13 @@ def test_saving_unicode_type_as_key(self):
...
         self.assertEqual(self.redis.get(u'Ñandu'), b'foo')
 
+    @pytest.importorskip('builtins.bytes', reason='future.types not available')
     def test_future_newbytes(self):
-        try:
-            from builtins import bytes
-        except ImportError:
-            pytest.skip('future.types not available')
         self.redis.set(bytes(b'\xc3\x91andu'), 'foo')
         self.assertEqual(self.redis.get(u'Ñandu'), b'foo')
```
bmerry: I'm not an expert on pytest, but I suspect that this change breaks the test. Firstly, the documentation for importorskip says that it takes a module, not a member of a module. It's also not documented to be a decorator. Finally, it probably still won't put `bytes` into the local scope.

I haven't tested it, but I think the solution is to put something like this into the body of the function (in place of the code you've removed).
```python
bytes = pytest.importorskip('builtins', reason='future.types not available').bytes
```

adamantike: You are right! `pytest.importorskip` can't be used as a decorator, because it returns the module it tries to import, when it's successful. Should be fixed now!

requirements.txt
adamantike: I can update it in this PR to maintain consistency, but repeatability could be harmful when working on shared libraries. For example, the project requires `sortedcontainers` to be installed, but it only tests `sortedcontainers==2.1.0` because that's what's in `requirements.txt`.

To fix it, we can completely migrate Travis to use Tox environments, and use something like this:
```
[tox]
envlist =
    py{27,34,35,36,37,py}-redis{210,30,31,32,33}

[testenv]
usedevelop = true
commands = pytest -v {posargs}
extras =
    lua
    testing
deps =
    redis210: redis==2.10.6
    redis30: redis==3.0.1
    redis31: redis==3.1.0
    redis32: redis==3.2.1
    redis33: redis==3.3.7
```

Also pinning any other library that could be needed. If tests start to fail because a dependency was updated, it means it's not correctly pinned in the `setup.py` file.

bmerry: It's impractical to test every possible combination of libraries, but at some stage I do want to add another item to the Travis build matrix that will test with the current version of redis, and that could also test with current versions of everything else too. I tend not to use tox much myself so I don't want to try to recreate what I've done with requirements.txt inside tox, and strict pinning in setup.py is not user-friendly as users may have a reason to use a different library version.

adamantike: I have updated this file by using the output from `pip freeze`, and merging versions for Python 2.7 and 3.7. It's clear how impractical it is to maintain this file when the project must work with multiple Python versions, and dependencies need multiple sub-dependencies, but what's how it works today. I can open a separate PR to show how it could work with `setup.py` and Tox.

> strict pinning in setup.py is not user-friendly as users may have a reason to use a different library version.
That's why I specified a `testing` extra, which is only used to test this project, and won't be needed by end users. For them, dependencies will be relaxed, to avoid forcing a particular version on their environments.

bmerry: > I have updated this file by using the output from pip freeze, and merging versions for Python 2.7 and 3.7.

Thanks! I hadn't realised just how much stuff pytest sucks in. I'll merge this now. I'm planning to drop Python 2 support at the end of the year (when Python 2 hits end-of-life) so things will get simpler then.

> I can open a separate PR to show how it could work with setup.py and Tox.

I appreciate the enthusiasm, but I'm happy with the way it works now and how it matches my workflow, so I probably wouldn't merge such a PR.

Issue #232, opened 2019-01-15
Support flags in ZADD


bmerry:
ZADD supports flags NX, XX, CH and INCR, which are not yet implemented.

PR #247, opened 2019-08-04
Support NX/XX/CH flags in ZADD command

adamantike:
Add support for the new `NX`, `XX`, and `CH` flags in the `ZADD` command, which can only be provided with `redis-py` version 3 installed.

This fixes a subset of what's filed in #232. `INCR` flag will be implemented in a different pull request, as it's completely different to handle.

### DIFF
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -1933,14 +1933,11 @@ class FakeSocket(object):
         ]
         old_len = len(zset)
         changed_items = 0
-        # Cache value to use on each iteration
-        not_nx_nor_xx = not nx and not xx
 
         for item_score, item_name in items:
             if (
-                not_nx_nor_xx
-                or (nx and item_name not in zset)
-                or (xx and item_name in zset)
+                (not nx or item_name not in zset)
+                and (not xx or item_name in zset)
             ):
                 if zset.add(item_name, item_score):
                     changed_items += 1
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -47,6 +47,7 @@ INVALID_SORT_FLOAT_MSG = "One or more scores can't be converted into double"
 SRC_DST_SAME_MSG = "source and destination objects are the same"
 NO_KEY_MSG = "no such key"
 INDEX_ERROR_MSG = "index out of range"
+ZADD_NX_XX_ERROR_MSG = "ZADD allows either 'nx' or 'xx', not both"
 ZUNIONSTORE_KEYS_MSG = "at least 1 input key is needed for ZUNIONSTORE/ZINTERSTORE"
 WRONG_ARGS_MSG = "wrong number of arguments for '{}' command"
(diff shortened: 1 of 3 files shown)


fakeredis/_server.py
```diff
@@ -1897,19 +1898,58 @@ def _apply_withscores(items, withscores):
...
+        changed_items = 0
+        # Cache value to use on each iteration
+        not_nx_nor_xx = not nx and not xx
+
+        for item_score, item_name in items:
+            if (
+                not_nx_nor_xx
+                or (nx and item_name not in zset)
+                or (xx and item_name in zset)
+            ):
```
bmerry: I think this condition can be simplified to `if (not nx or item_name not in zset) and (not xx or item_name in zset)`.

test_fakeredis.py
```diff
@@ -1797,6 +1797,64 @@ def test_zadd_multiple(self):
...
+
+    @redis3_only
+    def test_zadd_with_xx(self):
+        self.zadd('foo', {'four': 4, 'three': 3})
+        self.assertEqual(self.zadd('foo', {'four': -4, 'three': -3}, xx=True), 0)
+        self.assertEqual(self.redis.zrange('foo', 0, -1), [b'four', b'three'])
+        self.assertEqual(self.zadd('foo', {'four': 4, 'three': 3, 'zero': 0}, xx=True), 0)
+        self.assertEqual(self.redis.zrange('foo', 0, -1), [b'three', b'four'])
+        self.assertEqual(self.zadd('foo', {'two': 2, 'one': 1}, xx=True), 0)
+        self.assertEqual(self.redis.zrange('foo', 0, -1), [b'three', b'four'])
```
bmerry: Add a check that the scores do get updated.

test_fakeredis.py
```diff
@@ -1797,6 +1797,64 @@ def test_zadd_multiple(self):
         self.assertEqual(self.redis.zrange('foo', 1, 1),
                          [b'two'])
 
+    @redis3_only
+    def test_zadd_with_nx(self):
+        self.zadd('foo', {'four': 4, 'three': 3})
+        self.assertEqual(self.zadd('foo', {'four': 2, 'three': 1}, nx=True), 0)
```
bmerry: Add a  check that the score for 'four' doesn't get updated.

fakeredis/_server.py
adamantike: Fixed!

test_fakeredis.py
adamantike: I refactored the tests to use `zrange` with scores, and made them a bit more programmatic, to improve readability. Let me know what you think!

bmerry: I'm happy with that approach. What do you think about using a `namedtuple` rather than a dict with fixed keys to store the update information?

adamantike: Updated to use namedtuples, thanks for the suggestion! Let me know if naming works for you :)

bmerry:
Looks good. Thanks for the contribution!

PR #250, opened 2019-08-14
implement unlink command

zsblevins:
Note that a separate `_delete` function was written due to the attr manipulation done on `del_` which didn't allow for it to be called directly within unlink.


zsblevins:
Sure thing. I haven't used hypothesis before, so let me know if there's anything additional needed!

Issue #251, opened 2019-08-24
zscan gives error


abdolsamad:
When using redis, following code:
```python
import redis
rds = redis.StrictRedis(host="redis",port= 6379)
rds.zadd('key',{'val:0':1})
result = rds.zscan('key')
print(result)
```
produces:
```python
(0, [(b'val:0', 1.0)])
```

But in fake redis:
```python
import fakeredis
rds = fakeredis.FakeStrictRedis()
rds.zadd('key',{'val:0':1})
result = rds.zscan('key')
```
raises an exception in `zscan`:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/IPython/core/interactiveshell.py", line 3326, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-10-4814d682e624>", line 2, in <module>
    result = rds.zscan('key')
  File "/usr/local/lib/python3.7/site-packages/redis/client.py", line 1921, in zscan
    return self.execute_command('ZSCAN', *pieces, **options)
  File "/usr/local/lib/python3.7/site-packages/redis/client.py", line 838, in execute_command
    conn.send_command(*args)
  File "/usr/local/lib/python3.7/site-packages/redis/connection.py", line 670, in send_command
    check_health=kwargs.get('check_health', True))
  File "/usr/local/lib/python3.7/site-packages/redis/connection.py", line 649, in send_packed_command
    self._sock.sendall(item)
  File "/usr/local/lib/python3.7/site-packages/fakeredis/_server.py", line 777, in sendall
    self._parser.send(data)
  File "/usr/local/lib/python3.7/site-packages/fakeredis/_server.py", line 698, in _parse_commands
    self._process_command(fields)
  File "/usr/local/lib/python3.7/site-packages/fakeredis/_server.py", line 796, in _process_command
    result = self._run_command(func, sig, fields[1:], False)
  File "/usr/local/lib/python3.7/site-packages/fakeredis/_server.py", line 714, in _run_command
    result = func(*args)
  File "/usr/local/lib/python3.7/site-packages/fakeredis/_server.py", line 2056, in zscan
    return self._scan(key.value.items, cursor, *args)
  File "/usr/local/lib/python3.7/site-packages/fakeredis/_server.py", line 863, in _scan
    if cursor >= len(keys):
TypeError: object of type 'method' has no len()
```

bmerry:
[...]

PR #252, opened 2019-08-26
Fix implementation of ZSCAN

bmerry:
It still lacks unit tests. See #251.


bmerry:
@abdolsamad give this a try. I'll need to add a unit test before I merge, but it looks like it works for the case you were trying.

PR #257, opened 2019-11-23
Add other inputs for redis connection

### LINKED ISSUE #255: fakeredis does not support all the redis connection arguments
Some unsupported arguments include socket_timeout, socket_connect_timeout, socket_keepalive, socket_keepalive_options, socket_type, retry_on_timeout, socket_read_size and health_check_interval. While I wouldn't expect this to necessarily be simulated by fakeredis it would be nice to be able to use this in the from_url command.

Ex: FakeRedis.from_url(redis_url, db=None, socket_connect_timeout=5, socket_timeout=30)

nateyoder:
Closes #255

### DIFF
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -2490,12 +2490,22 @@ class FakeConnection(redis.Connection):
     description_format = "FakeConnection<db=%(db)s>"
 
     def __init__(self, server, db=0, password=None,
+                 socket_timeout=None, socket_connect_timeout=None,
+                 socket_keepalive=False, socket_keepalive_options=None,
+                 socket_type=0, retry_on_timeout=False,
                  encoding='utf-8', encoding_errors='strict',
                  decode_responses=False,
                  health_check_interval=0):
         self.pid = os.getpid()
         self.db = db
         self.password = password
+        # Allow socket attributes to be passed in and saved  even if they aren't used
+        self.socket_timeout = socket_timeout
+        self.socket_connect_timeout = socket_connect_timeout or socket_timeout
+        self.socket_keepalive = socket_keepalive
+        self.socket_keepalive_options = socket_keepalive_options or {}
+        self.socket_type = socket_type
+        self.retry_on_timeout = retry_on_timeout
         self.encoder = redis.connection.Encoder(encoding, encoding_errors, decode_responses)
         self.retry_on_timeout = False
         self._description_args = {'db': self.db}
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -2448,6 +2448,9 @@ delattr(FakeSocket, 'exec_')
 
 
 class _DummyParser(object):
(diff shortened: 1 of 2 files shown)


fakeredis/_server.py
```diff
@@ -2490,12 +2490,22 @@ class FakeConnection(redis.Connection):
...
+                 socket_timeout=None, socket_connect_timeout=None,
+                 socket_keepalive=False, socket_keepalive_options=None,
+                 socket_type=0, retry_on_timeout=False,
                  encoding='utf-8', encoding_errors='strict',
                  decode_responses=False,
                  health_check_interval=0):
         self.pid = os.getpid()
         self.db = db
         self.password = password
+        # Allow socket attributes to be passed in and saved  even if they aren't used
```
bmerry:
```suggestion
        # Allow socket attributes to be passed in and saved even if they aren't used
```

fakeredis/_server.py
```diff
@@ -2490,12 +2490,22 @@ class FakeConnection(redis.Connection):
     description_format = "FakeConnection<db=%(db)s>"
 
     def __init__(self, server, db=0, password=None,
+                 socket_timeout=None, socket_connect_timeout=None,
+                 socket_keepalive=False, socket_keepalive_options=None,
+                 socket_type=0, retry_on_timeout=False,
```
bmerry: Are these in the same position/order that the redis `__init__` takes them?

fakeredis/_server.py
```diff
@@ -2490,12 +2490,22 @@ class FakeConnection(redis.Connection):
...
                  decode_responses=False,
                  health_check_interval=0):
         self.pid = os.getpid()
         self.db = db
         self.password = password
+        # Allow socket attributes to be passed in and saved  even if they aren't used
+        self.socket_timeout = socket_timeout
+        self.socket_connect_timeout = socket_connect_timeout or socket_timeout
+        self.socket_keepalive = socket_keepalive
+        self.socket_keepalive_options = socket_keepalive_options or {}
```
bmerry: I take it this is equivalent to what real redis is setting on `self`?

fakeredis/_server.py
nateyoder: Yeah [at least as it is currently on master](https://github.com/andymccurdy/redis-py/blob/656b551eb247c491976646092a71320407c0fa03/redis/connection.py#L488).  I had forgotten a couple and just added them. Sorry.

fakeredis/_server.py
nateyoder: Yeah this was taken directly from [redis-py](https://github.com/andymccurdy/redis-py/blob/656b551eb247c491976646092a71320407c0fa03/redis/connection.py#L503).

bmerry:
Thanks, I'm going to merge. Don't worry about coveralls - it was also claiming a reduction in coverage on another PR that only touched the README, so I don't believe in it.

PR #258, opened 2019-11-28
Add note on how to get default server

### LINKED ISSUE #256: Add note on how to get default server
I use pytest and mock out redis in a fixture. One of the handy things fakeredis does is provide a way to programmatically mock connection errors. I think it would nice to note how to get the fake server instance in the case where you don't provide a server yourself. I can also do a PR for this is wanted, if not feel free to close this issue.

penguindustin:
Closes #256

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -92,6 +92,14 @@ your error handling. Simply set the connected attribute of the server to
   >>> r.set('foo', 'bar')
   True
 
+Or you can get the server reference directly with:
+
+.. code-block:: python
+
+  >>> import fakeredis
+  >>> r = fakeredis.FakeStrictRedis()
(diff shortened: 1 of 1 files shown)


bmerry:
Thanks for the suggestion. It's clearer now, but unfortunately this isn't something I want to recommend people do: `connection_kwargs` is an undocumented member of `redis.ConnectionPool`, so this code could be broken by a redis-py update. It wouldn't be the first time redis-py has made these sorts of internal changes. Such an update would of course break fakeredis, but that only has to be fixed once rather than fixed in every package that uses fakeredis this way.

penguindustin:
In that case, it totally makes sense not to add it here. Thanks for taking the time to listen to this suggestion!

PR #264, opened 2020-02-23
Added __call__ to make class as callable

shubham-sudo:
While mocking a redis connection I faced an issue "FakeStrictRedis is not callable".

```
@mock.patch(app.views.get_redis_connection, new_callable=FakeStrictRedis)
```


bmerry:
I'm not quite sure why that mock code doesn't work for you, but this is not the right fix. Instances of StrictRedis are not callable, so nor should instances of FakeStrictRedis be.

Issue #265, opened 2020-03-05
redis.log call in Lua script fails


upcFrost:
Hi,

Lua scripts fail for me if they use the `redis.log` directive. 

Example:
```
def test_redis_lua(self):
        lua = """redis.log(redis.LOG_WARNING, "foo bar")"""
        script = self.redis.register_script(lua)
        script()
```

Error:
```
.......
except LuaError as ex:
>           raise redis.ResponseError(str(ex))
E           redis.exceptions.ResponseError: [string "<python>"]:1: attempt to call a nil value (field 'log')
E           stack traceback:
E           	[string "<python>"]:1: in main chunk

../../venv/3/lib/python3.7/site-packages/fakeredis/_server.py:2352: ResponseError
```

I'm using fakeredis-1.2.1.

Adding a very basic stub like `redis.log = function(lvl, msg) return nil end` to `_server.py:2333` works for me, but ofc it'd be nice to log the messages

bmerry:
I'm afraid I don't use Lua myself, so I rely on PRs from users to extend the Lua functionality. If you have an idea of how log messages should be mapped to the Python logging system then please feel free to submit a PR.

upcFrost:
Ok, I'll try to make one 👌

bmerry:
Closed via #266.

PR #266, opened 2020-03-08
Lua redis.log implementation

upcFrost:
Added a very simple `redis.log` Lua call support, wired to the standard python `logging` lib.

Issue #265

### DIFF
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -14,7 +14,6 @@ import queue
 from collections import defaultdict
 from collections.abc import MutableMapping
 
-import lupa
 import six
 import redis
 
@@ -23,10 +22,16 @@ from ._zset import ZSet
 
 LOGGER = logging.getLogger('fakeredis')
 REDIS_LOG_LEVELS = {
-    b'LOG_DEBUG': logging.DEBUG,
-    b'LOG_VERBOSE': logging.INFO,
-    b'LOG_NOTICE': logging.INFO,
-    b'LOG_WARNING': logging.WARNING
+    b'LOG_DEBUG': 0,
+    b'LOG_VERBOSE': 1,
+    b'LOG_NOTICE': 2,
+    b'LOG_WARNING': 3
+}
+REDIS_LOG_LEVELS_TO_LOGGING = {
+    0: logging.DEBUG,
+    1: logging.INFO,
+    2: logging.INFO,
+    3: logging.WARNING
 }
 
 MAX_STRING_SIZE = 512 * 1024 * 1024
@@ -69,6 +74,9 @@ BAD_SUBCOMMAND_MSG = "Unknown {} subcommand or wrong # of args."
 BAD_COMMAND_IN_PUBSUB_MSG = \
     "only (P)SUBSCRIBE / (P)UNSUBSCRIBE / PING / QUIT allowed in this context"
 CONNECTION_ERROR_MSG = "FakeRedis is emulating a connection error."
+REQUIRES_MORE_ARGS = "{} requires {} arguments or more."
+LOG_WRONG_FIRST_ARG = "First argument must be a number (log level)."
+LOG_INVALID_DEBUG_LEVEL = "Invalid debug level."
 
 FLAG_NO_SCRIPT = 's'      # Command not allowed in scripts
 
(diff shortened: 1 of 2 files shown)


fakeredis/_server.py
```diff
@@ -2312,6 +2322,12 @@ def _lua_redis_pcall(self, lua_runtime, expected_globals, op, *args):
         except Exception as ex:
             return lua_runtime.table_from({b"err": str(ex)})
 
+    def _lua_redis_log(self, lua_runtime, expected_globals, lvl, *args):
+        self._check_for_lua_globals(lua_runtime, expected_globals)
+        if len(args) == 1:
+            msg = args[0]
+            LOGGER.log(lvl or logging.INFO, msg.decode('utf-8') if isinstance(msg, bytes) else msg)
```
bmerry: What does real redis do if the message is not a string (e.g. is an integer)?

test_fakeredis.py
```diff
@@ -4160,6 +4160,24 @@ def test_script(self):
         result = script(args=[42])
         self.assertEqual(result, b'42')
 
+    def test_lua_log(self):
+        if not isinstance(self.redis, fakeredis.FakeRedis) and not isinstance(self.redis, fakeredis.FakeStrictRedis):
```
bmerry:
```suggestion
        if not isinstance(self.redis, (fakeredis.FakeRedis, fakeredis.FakeStrictRedis)):
```

test_fakeredis.py
```diff
@@ -4160,6 +4160,24 @@ def test_script(self):
...
+            redis.log(redis.LOG_WARNING, "warning")
+        """
+        script = self.redis.register_script(script)
+        with self.assertLogs(logger, 'DEBUG') as cm:
+            script()
+            self.assertEqual(cm.output, ['DEBUG:%s:debug' % logger.name,
+                                         'INFO:%s:verbose' % logger.name,
+                                         'INFO:%s:notice' % logger.name,
+                                         'WARNING:%s:warning' % logger.name])
+
```
bmerry: Can you add another test (or several) which makes malformed calls e.g.
- Too many arguments
- Too few arguments (or no arguments)
- Message is not a string
- Level is not redis.LOG_*

fakeredis/_server.py
```diff
@@ -13,12 +14,21 @@
 from collections import defaultdict
 from collections.abc import MutableMapping
 
+import lupa
```
bmerry: Please remove this and import lupa only where it's needed. It's an optional dependency and fakeredis should work without it as long as Lua functionality isn't needed.

upcFrost: Removed

test_fakeredis.py
upcFrost: Fixed

upcFrost:
Added tests and fixed multiarg calls and type conversion. Redis actually does some strange things with types. String, int and float are all converted to string, while boolean is just ignored.

fakeredis/_server.py
```diff
@@ -23,10 +22,16 @@
...
+    b'LOG_VERBOSE': 1,
+    b'LOG_NOTICE': 2,
+    b'LOG_WARNING': 3
+}
+REDIS_LOG_LEVELS_TO_LOGGING = {
+    0: logging.DEBUG,
+    1: logging.INFO,
+    2: logging.INFO,
+    3: logging.WARNING
 }
```
bmerry: Is the double indirection because redis.LOG_DEBUG has the value 0 in real redis?

fakeredis/_server.py
```diff
@@ -2324,13 +2332,18 @@ def _lua_redis_pcall(self, lua_runtime, expected_globals, op, *args):
...
-            msg = args[0]
-            LOGGER.log(lvl or logging.INFO, msg.decode('utf-8') if isinstance(msg, bytes) else msg)
+        if len(args) < 1:
+            raise redis.ResponseError(REQUIRES_MORE_ARGS.format("redis.log()", "two"))
+        if lvl not in REDIS_LOG_LEVELS.values():
+            raise redis.ResponseError(LOG_INVALID_DEBUG_LEVEL)
+        msg = ' '.join([x.decode('utf-8')
+                        if isinstance(x, bytes) else str(x)
+                        for x in args if not isinstance(x, bool)])
+        LOGGER.log(REDIS_LOG_LEVELS_TO_LOGGING.get(lvl, logging.WARNING), msg)
```
bmerry: Is it necessary to use `get` rather than just `[]`, given that you've already checked that `lvl` is in `REDIS_LOG_LEVELS.values()`?

test_fakeredis.py
```diff
@@ -4178,6 +4178,45 @@ def test_lua_log(self):
                                          'INFO:%s:notice' % logger.name,
                                          'WARNING:%s:warning' % logger.name])
 
+    def test_lua_log_no_message(self):
+        if not isinstance(self.redis, (fakeredis.FakeRedis, fakeredis.FakeStrictRedis)):
+            pytest.skip("Works only on fakeredis, as testing real redis will require access to its logs")
```
bmerry: Is this skip necessary? Won't real redis raise the same ResponseError? Same applies to the other tests that expect a ResponseError.

test_fakeredis.py
```diff
@@ -4178,6 +4178,45 @@ def test_lua_log(self):
...
+        script = "redis.log(10, 'string')"
+        script = self.redis.register_script(script)
+        with self.assertRaises(redis.ResponseError):
+            script()
+
+    def test_lua_log_defined_vars(self):
+        if not isinstance(self.redis, (fakeredis.FakeRedis, fakeredis.FakeStrictRedis)):
+            pytest.skip("Works only on fakeredis, as testing real redis will require access to its logs")
```
bmerry: Since this code gets copy-pasted, it would be nice if it could be wrapped up into a decorator (so that it's stylistically similar to `@redis3_only` for example).

fakeredis/_server.py
upcFrost: Yes, and because user can actually use numbers instead of `redis.LOG_*` variables, just like in `logging`, but level values are different

PR #271, opened 2020-07-08
Make unit tests more pytest-onic

bmerry:
I wouldn't try to review everything - there are a *lot* of mechanical changes. Interesting bits are the smaller files and anything with pytest decorators (`@pytest.fixture`, `@pytest.mark`).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -331,7 +331,7 @@ bugs in Github.
    generally not produce the same results, and in Python versions before 3.6
    may produce different results each time the process is re-run.
 
-7. SCAN/ZSCAN/HSCAN/SSCAN will not necessary iterate all items if items are
+7. SCAN/ZSCAN/HSCAN/SSCAN will not necessarily iterate all items if items are
    deleted or renamed during iteration. They also won't necessarily iterate in
    the same chunk sizes or the same order as redis.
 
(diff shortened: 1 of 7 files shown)


bmerry:
The Travis failure seems to just be a transient thing where the test was too slow.

bmerry:
Oh, and you might want to review with whitespace differences ignored, because most tests got a dedent.

test/test_fakeredis.py
```diff
@@ -25,4529 +21,4692 @@
...
+    assert r.get('foo') == 'Ñandu'.encode()
+
+
+def test_saving_non_ascii_chars_as_key(r):
+    assert r.set('Ñandu', 'foo') is True
+    assert r.get('Ñandu') == b'foo'
+
+
+def test_saving_unicode_type_as_key(r):
+    assert r.set('Ñandu', 'foo') is True
```
ludwigschwardt, line 168: Should this not be `Ñandú`? 😂

README.rst
```diff
@@ -355,43 +355,38 @@ To ensure parity with the real redis, there are a set of integration tests
 that mirror the unittests.  For every unittest that is written, the same
 test is run against a real redis instance using a real redis-py client
 instance.  In order to run these tests you must have a redis server running
-on localhost, port 6379 (the default settings).  The integration tests use
-db=10 in order to minimize collisions with an existing redis instance.
+on localhost, port 6379 (the default settings). **WARNING**: the tests will
+completely wipe your database!
```
ludwigschwardt, line 359: Are you still using `db=10`? I guess a warning is in order in any case.

test/test_fakeredis.py
```diff
@@ -25,4529 +21,4692 @@
...
+            return cls(db=db, decode_responses=decode_responses, server=fake_server)
+        else:
+            cls = getattr(redis, name)
+            return cls('localhost', port=6379, db=db, decode_responses=decode_responses)
+
+    return factory
+
+
+@pytest.fixture
+def r(request, create_redis):
```
ludwigschwardt, line 104: Would the `r` fixture run before each test or is it cached somehow like `self.redis` used to be?

test/test_fakeredis.py
bmerry: No idea, it's code I inherited.

README.rst
bmerry: No, it hasn't used db=10 for a long time - and some tests use multiple databases.

test/test_fakeredis.py
bmerry: It runs for each test. I don't think there was any caching before though - what caching are you referring to?

test/test_fakeredis.py
ludwigschwardt: This is a Ñandú: https://en.wikipedia.org/wiki/Rhea_(bird)

Just wondered out loud 😄

test/test_fakeredis.py
ludwigschwardt: Ah right, I was thinking of `self.redis` stored on the test object but that is recreated in `setup` before each test anyway. Forgot about that...

PR #272, opened 2020-07-14
Fix coverage reports on coveralls.io

bmerry:
Pytest + coverage tends to work badly without an editable install, but in general it's preferable to run unit tests against a standard install since that's what users use. So make one case use an editable install for coverage purposes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,4 +37,5 @@ notifications:
   email:
     - js@jamesls.com
     - bmerry@ska.ac.za
-after_success: coveralls
+after_success:
+  - if [ "$COVERAGE" = "yes" ]; then coveralls; fi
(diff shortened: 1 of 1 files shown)


bmerry:
> Interesting, did not know that!

More details at https://github.com/pytest-dev/pytest-cov/issues/98. It works better when you have the source in a `src` subdirectory (which is why I've made katsdpmodels use that), but then coverage reports have long absolute pathnames.

ludwigschwardt:
A few people on that thread suggested adding an `__init__.py` to the `test` directory. See also [here](http://thomas-cokelaer.info/blog/2017/01/pytest-cov-collects-no-data-on-travis/). I mention it because fakeredis also lacks one.

bmerry:
> A few people on that thread suggested adding __init__.py to the test directory

I think the effect of that is to make pytest add the parent of the test directory to PYTHONPATH, thus causing it to pick up the in-tree version of the package instead of the installed version.

PR #277, opened 2020-08-14
Implement SET with KEEPTTL

mgetka:
Add support to `KEEPTTL` flag in the `SET` command which become available in `redis-py` 3.5.0.

### DIFF
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -1409,6 +1409,7 @@ class FakeSocket:
         px = None
         xx = False
         nx = False
+        keepttl = False
         while i < len(args):
             if casematch(args[i], b'nx'):
                 nx = True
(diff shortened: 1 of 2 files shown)


fakeredis/_server.py
```diff
@@ -1426,16 +1427,22 @@ def set_(self, key, value, *args):
                 if px <= 0:
                     raise SimpleError(INVALID_EXPIRE_MSG.format('set'))
                 i += 2
+            elif casematch(args[i], b'keepttl'):
+                keepttl = True
+                i += 1
             else:
                 raise SimpleError(SYNTAX_ERROR_MSG)
-        if (xx and nx) or (px is not None and ex is not None):
+        if (xx and nx) or ((px is not None) + (ex is not None) + keepttl > 1):
```
mgetka, line 1435: [`SET`](https://redis.io/commands/set) command documentation suggests that `KEEPTTL` option can be set next to `PX` or `EX` options, but in reality redis denies such requests. This behavior seems reasonable so I believe this is an error in the documentation and thus I enforce the actual (not the documented) behavior on the case of conflicting expiration options.

bmerry:
Thanks, good work!

PR #280, opened 2020-09-04
fixing documentation about previous singleton argument

galonsky:
I think this documentation is wrong. The new behavior is the same as setting `singleton=False` because that's the flag needed to NOT share data across instances, right?

Looked at the old documentation:

> Alternatively, you can create an instance that does not share data with other instances, by passing singleton=False to the constructor.


rbusquet (approved):

bmerry:
Thanks for catching this!

PR #286, opened 2020-11-20
Add support for TYPE option to SCAN command.

emorozov:
In Redis 6.0 there's a new option for the SCAN command that allows to limit types of returned keys. This PR adds this option.

### DIFF
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -878,7 +878,7 @@ class FakeSocket:
         However, provided the database is not modified, every key will be
         returned exactly once.
         """
-        pattern = None
+        pattern = type = None
         count = 10
         if len(args) % 2 != 0:
             raise SimpleError(SYNTAX_ERROR_MSG)
@@ -889,6 +889,16 @@ class FakeSocket:
                 count = Int.decode(args[i + 1])
                 if count <= 0:
                     raise SimpleError(SYNTAX_ERROR_MSG)
+            elif casematch(args[i], b'type'):
+                type = args[i + 1]
+                is_valid = (casematch(type, b'string') or
+                            casematch(type, b'list') or
+                            casematch(type, b'set') or
+                            casematch(type, b'zset') or
+                            casematch(type, b'stream') or
+                            casematch(type, b'hash'))
+                if not is_valid:
+                    raise SimpleError(SYNTAX_ERROR_MSG)
             else:
                 raise SimpleError(SYNTAX_ERROR_MSG)
 
@@ -898,11 +908,16 @@ class FakeSocket:
         result_cursor = cursor + count
         result_data = []
 
+        match_key = match_type = lambda key: True
         if pattern is not None:
-            regex = compile_pattern(pattern)
+            match_key = compile_pattern(pattern).match
+        if type is not None:
+            match_type = lambda key: self.type(self._db[key]).value == type
+
+        if pattern is not None or type is not None:
             for val in itertools.islice(data, cursor, result_cursor):
                 compare_val = val[0] if isinstance(val, tuple) else val
-                if regex.match(compare_val):
+                if match_key(compare_val) and match_type(compare_val):
                     result_data.append(val)
         else:
             result_data = data[cursor:result_cursor]
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -908,16 +908,18 @@ class FakeSocket:
         result_cursor = cursor + count
         result_data = []
 
-        match_key = match_type = lambda key: True
-        if pattern is not None:
-            match_key = compile_pattern(pattern).match
-        if type is not None:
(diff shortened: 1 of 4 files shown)


emorozov:
Some tests are failing because this change requires recent redis-py.

bmerry:
> Some tests are failing because this change requires recent redis-py.

You can use `@pytest.mark.skipif` to skip them e.g. [like this](https://github.com/jamesls/fakeredis/blob/e04fc6e24baaaceb582e95a3d61b63d34a9e634d/test/test_fakeredis.py#L683).

emorozov:
@bmerry Thank you. Updated PR, now all the tests pass.

fakeredis/_server.py
```diff
@@ -878,7 +878,7 @@ def _scan(self, keys, cursor, *args):
         However, provided the database is not modified, every key will be
         returned exactly once.
         """
-        pattern = None
+        pattern = type = None
```
bmerry, line 881: I'm not a fan of squashing multiple assignments onto one line like this.
```suggestion
        pattern = None
        type = None
```

fakeredis/_server.py
```diff
@@ -898,11 +908,18 @@ def _scan(self, keys, cursor, *args):
         result_cursor = cursor + count
         result_data = []
 
-        if pattern is not None:
-            regex = compile_pattern(pattern)
+        regex = compile_pattern(pattern) if pattern is not None else None
+
+        def _match_key(key):
```
bmerry, line 913: No need for the leading underscore - the name isn't accessible from outside the method.

fakeredis/_server.py
```diff
@@ -889,6 +889,16 @@ def _scan(self, keys, cursor, *args):
...
+            elif casematch(args[i], b'type'):
+                type = args[i + 1]
+                is_valid = (casematch(type, b'string') or
+                            casematch(type, b'list') or
+                            casematch(type, b'set') or
+                            casematch(type, b'zset') or
+                            casematch(type, b'stream') or
+                            casematch(type, b'hash'))
+                if not is_valid:
+                    raise SimpleError(SYNTAX_ERROR_MSG)
```
bmerry, line 901: Delete this check. The real redis server doesn't check it.

fakeredis/_server.py
```diff
@@ -898,11 +908,18 @@ def _scan(self, keys, cursor, *args):
...
 
-        if pattern is not None:
-            regex = compile_pattern(pattern)
+        regex = compile_pattern(pattern) if pattern is not None else None
+
+        def _match_key(key):
+            return regex.match(key) if pattern is not None else True
+
+        def _match_type(key):
+            return self.type(self._db[key]).value == type if type else True
```
bmerry, line 917: Needs to be a `casematch` (or else convert `type` to lowercase when it is loaded from args). Also since I've asked above for the value of `type` not to be checked, it could be the empty string and hence `if type` should become `if type is not None`.

test/test_fakeredis.py
```diff
@@ -1633,6 +1633,20 @@ def test_scan_iter_multiple_pages_with_match(r):
...
+@pytest.mark.min_server('6.0')
+def test_scan_iter_multiple_pages_with_type(r):
+    all_keys = key_val_dict(size=100)
+    assert all(r.set(k, v) for k, v in all_keys.items())
+    # Now add a few keys of another type
+    zadd(r, 'zset1', {'otherkey': 1})
+    zadd(r, 'zset2', {'andanother': 1})
+    actual = set(r.scan_iter(_type='string'))
+    assert actual == set(all_keys)
+    actual = set(r.scan_iter(_type='zset'))
```
bmerry, line 1646: Maybe use ZSET (uppercase) here to check that the match is case insensitive.

bmerry (changes requested):

fakeredis/_server.py
emorozov: Fixed.

fakeredis/_server.py
emorozov: Fixed.

fakeredis/_server.py
emorozov: Removed.

fakeredis/_server.py
emorozov: Fixed.

test/test_fakeredis.py
emorozov: Fixed.

emorozov:
Thank you for the review. I've implemented all the suggestions.

bmerry:
Thanks. It looks good now, so I'm just waiting for Travis to eventually get around to the testing (which seems to take a lot longer these days - probably time to migrate to Github Actions at some point).

bmerry (approved):

bmerry:
@ludwigschwardt FYI - I seem to recall you were interested in this functionality for katdal.

PR #289, opened 2020-12-04
Adding Power support(ppc64le) with ci and testing to the project for architecture independent

asellappen:
I am working for IBM to port cpu arch ppc64le for open sources.

This PR adds CI support for the IBM Power Little Endian (ppc64le) architecture. The idea is to ensure that the builds on this architecture are continuously tested along with the Intel builds (amd64) as this is part of the ubuntu distro 
on that architecture as well and detecting (and fixing) any issues or failures early would help to ensure that we are always up to date.

This helps us simplify testing later when distributions are re-building and re-releasing,We typically build applications for customers and 
ISVs, and while we don't use this package directly,we do count on all of the packages in debian/ubuntu to build other packages. So we more likely have this as a second or 
third level dependency and couldn't tell you explicitly which features we use or our usage model,For more info tag @gerrith3.

Please help to verify and merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,26 @@ jobs:
     - env: REDIS_PY="3.4.1"
     - env: REDIS_PY="3.5.0"
     - env: REDIS_PY="3.5.*" COVERAGE="yes"
+    - python: 3.6
+      arch: ppc64le
+    - python: 3.7
+      arch: ppc64le
+    - python: 3.8
+      arch: ppc64le
+    - env: REDIS_PY="3.0.1"
+      arch: ppc64le
+    - env: REDIS_PY="3.1.0"
+      arch: ppc64le
+    - env: REDIS_PY="3.2.1"
+      arch: ppc64le
+    - env: REDIS_PY="3.3.11"
+      arch: ppc64le
+    - env: REDIS_PY="3.4.1"
+      arch: ppc64le
+    - env: REDIS_PY="3.5.0"
+      arch: ppc64le
+    - env: REDIS_PY="3.5.*" COVERAGE="yes"
+      arch: ppc64le
 cache:
   - pip
 services:


bmerry:
This project is pure Python code and hence architecture-independent. I don't see that testing on more architectures adds any value, and it will slow down the unit tests (which already take longer than I would like).

asellappen:
thanks for taking look at PR, Even this project is pure Python code and architecture-independent, We are bit worrying about dependent tools like redis which is written in c .Hence we are trying to test for ppc64le and rollout.

and also regarding time taken for the run, It ran on travis-ci.org which servers are slow and they are making us to move to travis-ci.com.

bmerry:
> We are bit worrying about dependent tools like redis which is written in c

Surely then it would make more sense to test the dependencies directly? Redis for example has its own very comprehensive test suite which will cover parts of redis that fakeredis tests won't.

Anyway, fakeredis doesn't depend on redis to run. It only uses redis as a reference for the unit tests.

gerrith3:
@bmerry we've found that everything is a dependency for something else at some point in time or another.  ;)  That's a little tongue in cheek but my team basically picks up exploding bombs of software the fails on ppc64le based on what some user or another is doing, some new workload, whatever.  We are at the point where we are trying to bring ppc64le testing up to the point of Intel testing across the board to reduce the number of failures end users see.

And, while there are lots of potentially architecture independent languages, we've found that all of them can wind up with situations which depend on out of tree binaries (often Intel based), have callouts to custom amd64 code, use various operating system interfaces aka sys. APIs in python (several of which depend on the underlying hardware), and various other items.   One of the most common is differences in the build environment, which we currently rule out by having the team test and validate the current top of tree, and potentially fix the build environment if needed.  Any of these can get out of date, and the only way to really prove that it works is to build the project on actual ppc64le based hardware.  Hence our drive to get as many of these easier ones converted as possible.

bmerry:
I'm still not convinced, but let me see how the build times look once things are on travis-ci.com.

bmerry:
@jamesls are you able to do the migration to travis-ci.com? I've sent you an email with some details.

bmerry:
The CI has been moved to Github Actions, so I'm going to close this PR. If you still want this you can submit another PR against the Github Actions workflow, but I'll only consider it if it adds a single job to the matrix rather than doubling the matrix size.

PR #293, opened 2021-02-03
Align `FakeConnection` constructor signature to base class

### LINKED ISSUE #234: Failing to migrate to v1.0 on django project
This is on a django 1.11 / python 2.7 project. (also django-redis==4.10.0, redis==2.10.6)
I can't work out how to migrate my settings to be compatible with version 1.0. Or is the setup contradictory to what 1.0 needs?

This is how it worked with the <1.0:

```
In [1]: settings.CACHES
Out[1]: 
{'admin': {'BACKEND': 'django_redis.cache.RedisCache',
  'KEY_PREFIX': 'admin',
  'LOCATION': 'redis://redis:6379/3',
  'OPTIONS': {'CLIENT_CLASS': 'django_redis.client.DefaultClient',
   'REDIS_CLIENT_CLASS': 'fakeredis.FakeStrictRedis'}},
 'default': {'BACKEND': 'django.core.cache.backends.dummy.DummyCache',
  'LOCATION': 'foobar'},
 'sessions': {'BACKEND': 'django_redis.cache.RedisCache',
  'KEY_PREFIX': 'shop-sessions',
  'LOCATION': 'redis://redis:6379/2',
  'OPTIONS': {'CLIENT_CLASS': 'django_redis.client.DefaultClient',
   'REDIS_CLIENT_CLASS': 'fakeredis.FakeStrictRedis'}}}
```

With v1.0 I receive this on pretty much all my tests:

```
======================================================================
ERROR: test_something (foobar.messaging.tests.test_ticket_views.ViewsSupportingCORSTestCase)
----------------------------------------------------------------------
  File "/var/www/shop/current/app/django-sites/foobar/utils/testcases.py", line 603, in tearDown
    self.client.logout()
  File "/var/venv-stable/lib/python2.7/site-packages/django/test/client.py", line 685, in logout
    if self.session:
  File "/var/venv-stable/lib/python2.7/site-packages/django/test/client.py", line 459, in session
    session.save()
  File "/var/venv-stable/lib/python2.7/site-packages/django/contrib/sessions/backends/cache.py", line 57, in save
    return self.create()
  File "/var/venv-stable/lib/python2.7/site-packages/django/contrib/sessions/backends/cache.py", line 44, in create
    self._session_key = self._get_new_session_key()
  File "/var/venv-stable/lib/python2.7/site-packages/django/contrib/sessions/backends/base.py", line 164, in _get_new_session_key
    if not self.exists(session_key):
  File "/var/venv-stable/lib/python2.7/site-packages/django/contrib/sessions/backends/cache.py", line 71, in exists
    return session_key and (self.cache_key_prefix + session_key) in self._cache
  File "/var/venv-stable/lib/python2.7/site-packages/django/core/cache/backends/base.py", line 203, in __contains__
    return self.has_key(key)
  File "/var/venv-stable/lib/python2.7/site-packages/django_redis/cache.py", line 39, in _decorator
    raise e.parent
ConnectionError: Error -2 connecting to redis:6379. Name or service not known.

```

It appears to be looking for the "real" redis server still. 
If I change REDIS_CLIENT_CLASS to CLIENT_CLASS I get:

```
    self.client.logout()
  File "/var/venv-stable/lib/python2.7/site-packages/django/test/client.py", line 685, in logout
    if self.session:
  File "/var/venv-stable/lib/python2.7/site-packages/django/test/client.py", line 459, in session
    session.save()
  File "/var/venv-stable/lib/python2.7/site-packages/django/contrib/sessions/backends/cache.py", line 57, in save
    return self.create()
  File "/var/venv-stable/lib/python2.7/site-packages/django/contrib/sessions/backends/cache.py", line 44, in create
    self._session_key = self._get_new_session_key()
  File "/var/venv-stable/lib/python2.7/site-packages/django/contrib/sessions/backends/base.py", line 164, in _get_new_session_key
    if not self.exists(session_key):
  File "/var/venv-stable/lib/python2.7/site-packages/django/contrib/sessions/backends/cache.py", line 71, in exists
    return session_key and (self.cache_key_prefix + session_key) in self._cache
  File "/var/venv-stable/lib/python2.7/site-packages/django/core/cache/backends/base.py", line 203, in __contains__
    return self.has_key(key)
  File "/var/venv-stable/lib/python2.7/site-packages/django_redis/cache.py", line 32, in _decorator
    return method(self, *args, **kwargs)
  File "/var/venv-stable/lib/python2.7/site-packages/django_redis/cache.py", line 124, in has_key
    return self.client.has_key(*args, **kwargs)
AttributeError: 'FakeStrictRedis' object has no attribute 'has_key'


```

wandering-tales:
As `FakeConnection` is a subclass of the `Connection` class in `redis-py`, in order to honor OOP substitutability principle, the signature of the constructor should be aligned with the one of the base class. That means it should contain at least all the parameters of the base class.

This automatically fixes a problem occurring when the connection object is dynamically instantiated within the context of the default connection pool class in `redis-py`, named `ConnectionPool`, and when, in turn, the latter is dynamically instantiated using its standard `from_url` factory method. In such scenario the instantiation of the connection object fails because the `FakeConnection` class misses both `host` and `port` parameters in its constructor signature.

That happens because the `from_url` factory method above, given a canonical Redis URL, guesses several connection parameters values, among which `host`, `username`, `password`, `path`, and `db`, which values override any other corresponding value passed to the factory method itself via the connection variadic arguments.

Aligning the `FakeConnection` constructor signature to the one of its base class, by including the `host` and `port` parameters, is helpful in contexts where there's no direct control over the instantiantiation of Redis clients, connections and connection pools objects (e.g. `django-redis` library): whilst preserving current behavior, this avoids to workaround the issue by defining either custom connection factories or custom connection pools.

Closes #234

### DIFF
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -1,5 +1,4 @@
 import logging
-import os
 import time
 import threading
 import math
@@ -2655,39 +2654,10 @@ class FakeSelector(BaseSelector):
 class FakeConnection(redis.Connection):
     description_format = "FakeConnection<db=%(db)s>"
 
-    def __init__(self, server, db=0, username=None, password=None,
-                 socket_timeout=None, socket_connect_timeout=None,
-                 socket_keepalive=False, socket_keepalive_options=None,
-                 socket_type=0, retry_on_timeout=False,
-                 encoding='utf-8', encoding_errors='strict',
-                 decode_responses=False, parser_class=_DummyParser,
-                 socket_read_size=65536, health_check_interval=0,
-                 client_name=None):
-        self.pid = os.getpid()
-        self.db = db
-        self.username = username
-        self.client_name = client_name
-        self.password = password
-        # Allow socket attributes to be passed in and saved even if they aren't used
-        self.socket_timeout = socket_timeout
-        self.socket_connect_timeout = socket_connect_timeout or socket_timeout
-        self.socket_keepalive = socket_keepalive
-        self.socket_keepalive_options = socket_keepalive_options or {}
-        self.socket_type = socket_type
-        self.retry_on_timeout = retry_on_timeout
-        self.encoder = redis.connection.Encoder(encoding, encoding_errors, decode_responses)
-        self._description_args = {'db': self.db}
-        self._connect_callbacks = []
-        self._buffer_cutoff = 6000
-        self._server = server
-        # self._parser isn't used for anything, but some of the
-        # base class methods depend on it and it's easier not to
-        # override them.
-        self._parser = parser_class(socket_read_size=socket_read_size)
-        self._sock = None
-        # added in redis==3.3.0
-        self.health_check_interval = health_check_interval
-        self.next_health_check = 0
+    def __init__(self, *args, **kwargs):
+        self._server = kwargs.pop('server')
+        self._description_args = {'db': kwargs.get('db', 0)}
+        super().__init__(*args, **kwargs)
 
     def connect(self):
         super().connect()
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -2656,7 +2656,6 @@ class FakeConnection(redis.Connection):
 
     def __init__(self, *args, **kwargs):
         self._server = kwargs.pop('server')
-        self._description_args = {'db': kwargs.get('db', 0)}
         super().__init__(*args, **kwargs)
 
     def connect(self):
--- a/test/test_fakeredis.py
+++ b/test/test_fakeredis.py
@@ -4988,6 +4988,7 @@ class TestInitArgs:
         db.set('foo', 'bar')
         assert db.get('foo') == 'bar'
 
+    @redis3_only
     def test_can_allow_extra_args(self):
         db = fakeredis.FakeStrictRedis.from_url(
             'redis://localhost:6379/0',


wandering-tales:
My actual workaround to use _fakeredis_ to mock a Redis instance used as a cache backend via [django-redis](https://github.com/jazzband/django-redis) library in a Django project is the following:

```python
import django.core.cache

from django.test import TestCase, override_settings

import fakeredis
import mock

_caches = {
    'default': {
        'BACKEND': 'django_redis.cache.RedisCache',
        'LOCATION': 'redis://fake-redis/0',
        'OPTIONS': {
            'CLIENT_CLASS': 'django_redis.client.DefaultClient',
        }
    }
}

@override_settings(CACHES=_caches)
@mock.patch("django_redis.pool.ConnectionFactory.get_connection",
            new=mock.MagicMock(return_value=fakeredis.FakeRedis()))
class TestCache(TestCase):
    def test_set_unicode_key_redis_cache(self):
        try:
            django.core.cache.caches['default'].set('unicode-key-€', 'fake_value')
        except UnicodeEncodeError:
            self.fail("`RedisCache.set` raised "
                      "UnicodeEncodeError unexpectedly")
```

bmerry:
Thanks. This seems reasonable, and I think using the base class `__init__` should help make it more robust.

Travis is showing a flake8 error that is preventing the rest of the unit tests from running. I think what is likely to happen is that it will fail on older versions of redis-py that didn't have all the arguments that the current Connection class takes. It may be possible to fix that by having the constructor take `*args, **kwargs` instead of named arguments.

wandering-tales:
@bmerry Thanks to you for your quick reply.

I checked all the Travis CI builds and indeed the flake8 commands failed, but apparently for another reason:
```
$ flake8
./fakeredis/_server.py:2:1: F401 'os' imported but unused
The command "flake8" failed and exited with 1 during .
```
This is caused by the fact that I removed some code in the `__init__` that was already included in the related redis-py baseclass.

I removed that unused import. Let's see.

bmerry:
It looks like there are now some failures of the sort I expected, but also these:
```
E           assert WrappedException(ResponseError("wrong number of arguments for 'exec' command")) == WrappedException(ExecAbortError("Transaction discarded because of: wrong number of arguments for 'exec' command"))
```

That seems to be some bug on master that for some reason hasn't appeared before, so don't worry about those.

wandering-tales:
You're right. Tests actually fails for redis-py versions lower than 3.3.11.

I followed your suggestion to use variadic arguments. I think it's a safer way to guarantee compatibility with multiple redis-py interfaces.

wandering-tales:
@bmerry Excluding the test you mentioned, now only one the CI job using redis-py 2.10.6 fails. Precisely, only the following test fails:

`TestInitArgs.test_can_allow_extra_args`

At first glance, it's clear that the purpose of that test is to ensure the `FakeConnection` object can be instantiated with any type of arguments. Of course, now that the constructor of the base class `redis.Connection` is called, that is not possible anymore, as the parent object does not support variadic arguments. Nonetheless, I'm missing which particular usage scenario the test is trying to cover.

If I use fakeredis in a project where I use redis-py 2.10.6, I should be aware I cannot instantiate a `FakeStrictRedis` object with **any** parameter I want, as its interface is bound to that particular version of redis-py. What is the reason for allowing extra arguments to the client that are not supported by the signature of the base class, which signature depends on the specific version of redis-py? The only reason I can see is that in that project I need to be able to run code that uses fakeredis against multiple versions of redis-py. Am I missing something else?

bmerry:
I've fixed up master.

That test is intended to ensure that arguments that can be passed to redis can also be passed to fakeredis, even though it doesn't use them. I think the simplest thing is just to add the `@redis3_only` decorator to it.

wandering-tales:
>  I think the simplest thing is just to add the @redis3_only decorator to it.

That makes sense.

bmerry:
In future please don't rewrite history on a PR. It makes it impossible to see what changed since my previous review.

fakeredis/_server.py
```diff
@@ -2655,39 +2654,10 @@ def check_is_ready_for_command(self, timeout):
...
-        # base class methods depend on it and it's easier not to
-        # override them.
-        self._parser = parser_class(socket_read_size=socket_read_size)
-        self._sock = None
-        # added in redis==3.3.0
-        self.health_check_interval = health_check_interval
-        self.next_health_check = 0
+    def __init__(self, *args, **kwargs):
+        self._server = kwargs.pop('server')
+        self._description_args = {'db': kwargs.get('db', 0)}
```
bmerry, line 2659: Is this still necessary? It was there for older versions of redis (e.g. 2.10.6), but `super().__init__` should now take care of it (and in fact is probably overwriting it).

wandering-tales:
> In future please don't rewrite history on a PR. It makes it impossible to see what changed since my previous review.
Sorry, I won't do it again. I did because sometimes I found out that Travis CI do not automatically run Pull Request builds.

fakeredis/_server.py
wandering-tales: I see that the `FakeRedisMixin` constructor still accepts `server` parameter which is passed as a keyword to the Redis `ConnectionPool` class. In turn the `ConnectionPool` kwargs are passed to the constructor of the specific connection class.

Indeed the default Redis `Connection` class do not expect that parameter anymore in latest versions, but it seems it's needed by the `FakeConnection` class, precisely in the internal `_connect` method:

```python
def _connect(self):
        if not self._server.connected:
            raise redis.ConnectionError(CONNECTION_ERROR_MSG)
```

Regarding the `_description_args`, instead, I don't find any reference elsewhere. So it's probably a leftover.

bmerry: I was referring specifically to `_description_args`. Let's delete it.

wandering-tales:
I sure can.

PR #298, opened 2021-05-25
Defer the cleanup in FakeSocket.close

### LINKED ISSUE #297: Deadlock triggered by RQ with a scheduler
When running RQ with its scheduler enabled, we inevitably deadlock because a Connection object is collected while the main server lock is being held by the command processor. This reliably triggers it in a matter of seconds.

```python
import fakeredis
import rq

import footoo

connection = fakeredis.FakeStrictRedis()
queue = rq.Queue("Q", is_async=False, connection=connection)


while True:
    queue.enqueue(footoo.what)
    worker = rq.SimpleWorker(queues=[queue], connection=connection)
    worker.work(burst=True, with_scheduler=True)
```

(It doesn't matter what footoo.what does, it just needs to be outside the `__main__` module so RQ can refer to it).

It is caused because at some point a `Connection` object is being finalized and its `__del__` method calls the `FakeSocket.close` method which tries to acquire a lock that's already being held.

A very easy fix is to replace the `server.lock` with a `threading.RLock`. This doesn't cause any test failures but I'm not sure how thoroughly concurrency is tested. It seems likely to cause problems for people using cooperative threading models, like AIO, where multiple tasks running on the same thread will be permitted to acquire the RLock even though they should be exclusive.

bmerry:
It can be called from the garbage collector in the middle of other code
that is holding the server lock, so could deadlock. It now just appends
itself to a cleanup list, and the next command does the cleanup.

Closes #297.

### DIFF
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -274,7 +274,7 @@ class Database(MutableMapping):
     def __init__(self, lock, *args, **kwargs):
         self._dict = dict(*args, **kwargs)
         self.time = 0.0
-        self._watches = defaultdict(set)      # key to set of connections
+        self._watches = defaultdict(weakref.WeakSet)      # key to set of connections
         self.condition = threading.Condition(lock)
         self._change_callbacks = set()
 
@@ -640,6 +640,8 @@ class FakeServer:
         self.psubscribers = defaultdict(weakref.WeakSet)
         self.lastsave = int(time.time())
         self.connected = True
+        # List of weakrefs to sockets that are being closed lazily
+        self.closed_sockets = []
 
 
 class FakeSocket:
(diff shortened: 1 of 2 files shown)


detzgk:
Looks all good on my end, the previously failing script can now run for long periods of time without deadlocking.

fakeredis/_server.py
```diff
@@ -664,7 +666,12 @@ def __init__(self, server):
         self._parser.send(None)
 
     def put_response(self, msg):
-        self.responses.put(msg)
+        # redis.Connection.__del__ might call self.close at any time, which
+        # will set self.responses to None. We assume this will happen
+        # atomically, and the code below then protects us against this.
+        responses = self.responses
+        if responses:
+            self.responses.put(msg)
```
ludwigschwardt, line 674: Should this not be `responses.put(msg)`? Otherwise it could still hit `None` the second time around.

ludwigschwardt (commented):
Just the one question. The rest looks good.

fakeredis/_server.py
bmerry: Doh! Yes, that's what I had in my head.

bmerry: Fixed, thanks.

ludwigschwardt (approved):

PR #303, opened 2021-07-30
SINTER[STORE] requires keys to be sets

jspricke:
Starting with Redis 6.0.15 this behaviour changed.
The definition of SINTER[STORE] states:

"Keys that do not exist are considered to be empty sets."

At the same time SINTER only accepts set:

"intersection of all the given sets"

Both quotes from: https://redis.io/commands/sinter.

The behaviour of Redis 6.0.14 was that it ignored the type of later keys
if it found an empty set and returned that. Radis 6.0.15 changed this
behaviour to return a WRONGTYPE if it finds a non set key in the
arguments.

Example to reproduce:

127.0.0.1:6379> FLUSHALL
OK
127.0.0.1:6379> SINTER a b
(empty array)
127.0.0.1:6379> SET b something
OK
127.0.0.1:6379> SINTER a b
(error) WRONGTYPE Operation against a key holding the wrong kind of value

Cf. https://github.com/redis/redis/issues/9273.

### DIFF
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -1900,13 +1900,11 @@ class FakeSocket:
     def sdiffstore(self, dst, *keys):
         return self._setop(lambda a, b: a - b, False, dst, *keys)
 
-    # The following keys can't be marked as sets because of the
-    # stop_if_empty early-out.
-    @command((Key(set),), (Key(),))
+    @command((Key(set),), (Key(set),))
     def sinter(self, *keys):
         return self._setop(lambda a, b: a & b, True, None, *keys)
 
-    @command((Key(), Key(set)), (Key(),))
+    @command((Key(), Key(set)), (Key(set),))
     def sinterstore(self, dst, *keys):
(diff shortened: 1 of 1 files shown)


bmerry:
Thanks. Unfortunately with these subtle behaviour changes that Redis makes in patch-level updates, it's not possible for fakeredis to match the behaviour of every user's Redis server. At the moment it's targeting [6.0.10](https://github.com/jamesls/fakeredis/blob/f8b59367133ceff0f40aa5a3f6f88e851b8aff8e/.github/workflows/test.yml#L34), and the unit tests for this PR are all failing as a result. At some stage I'll get around to updating it (probably to 6.2.x) and deal with all the bugs that shakes out, at which point I'll take a look at this PR again. From a quick glance at the code it looks good.

bmerry:
Fakeredis is under new maintenance at https://github.com/dsoftwareinc/fakeredis-py. I am thus closing all PRs on this repository. If you feel your PR is still relevant, please open a new one on that repository.

PR #305, opened 2021-08-16
Packaging modernisation

bmerry:
- Add a pyproject.toml file (with setuptools_scm, so that pypa-build
  will include everything from SCM in the build; and remove MANIFEST.in)
- Move all the metadata from setup.py to setup.cfg (setup.py is left
  just to support editable builds).
- Use `attr:` in setup.cfg to get the version programmatically instead
  of repeating it.
- Replace distutils.version with packaging.version (the former will be deprecated by PEP 632).
- Update my email address

### DIFF
--- a/fakeredis/aioredis.py
+++ b/fakeredis/aioredis.py
@@ -1,9 +1,8 @@
-import distutils.version
-
 import aioredis
+import packaging.version
 
 
-if aioredis.__version__ >= distutils.version.StrictVersion('2.0.0a1'):
+if packaging.version.Version(aioredis.__version__) >= packaging.version.Version('2.0.0a1'):
     from ._aioredis2 import FakeConnection, FakeRedis  # noqa: F401
 else:
     from ._aioredis1 import (  # noqa: F401
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[build-system]
+requires = ["setuptools", "wheel", "setuptools-scm"]
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,6 +22,7 @@ classifiers =
 [options]
 packages = fakeredis
 install_requires =
+    packaging
     # Minor version updates to redis tend to break fakeredis. If you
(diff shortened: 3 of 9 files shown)


fakeredis/aioredis.py
```diff
@@ -1,9 +1,8 @@
-import distutils.version
-
 import aioredis
+import packaging.version
```
ludwigschwardt, line 2: I would probably use the same scheme as all the other files (`from packaging.version import Version`), but that's a very minor suggestion.

ludwigschwardt (commented):
I don't see a `pyproject.toml`...

setup.cfg
```diff
@@ -1,3 +1,43 @@
+[metadata]
+name = fakeredis
+version = attr: fakeredis.__version__
```
ludwigschwardt, line 3: Interesting! I did not know about these directives, even though it is probably very old :-D

fakeredis/aioredis.py
bmerry: I wanted to avoid exposing a `fakeredis.aioredis.Version`. I know I could import it as `_Version` or something but that still wouldn't be consistent.

pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[build-system]
+requires = ["setuptools", "wheel", "setuptools-scm"]
```
ludwigschwardt, line 2: Do you have to activate `setuptools_scm` beyond this line?

bmerry: It doesn't seem to need anything else: when I run `python -m build .` all the files end up in the sdist. I vaguely recall that the version handling needs activation, but I'm not using it.

fakeredis/aioredis.py
bmerry: It's a public module. `fakeredis._aioredis1` and `fakeredis._aioredis2` are internal.

pyproject.toml
ludwigschwardt: Makes sense.

ludwigschwardt (approved):

PR #309, opened 2021-10-05
Add support for `zpopmin`

eliangcs:
Matches the `zpopmin(key, count=None)` function signature in [redis-py](https://redis-py.readthedocs.io/en/stable/#redis.Redis.zpopmin). I could add `zpopmax` while I was there but that's all we need for now.

### DIFF
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -2013,7 +2013,7 @@ class FakeSocket:
         return OK
 
     # Sorted set commands
-    # TODO: [b]zpopmin/zpopmax,
+    # TODO: bzpopmin/[b]zpopmax,
 
     @staticmethod
     def _limit_items(items, offset, count):
@@ -2128,6 +2128,20 @@ class FakeSocket:
     @command((Key(ZSet), StringTest, StringTest))
     def zlexcount(self, key, min, max):
         return key.value.zlexcount(min.value, min.exclusive, max.value, max.exclusive)
+
+    @command((Key(ZSet),), (Int,))
+    def zpopmin(self, key, count=None):
+        zset = key.value
+        min_score = ScoreTest.decode(b'-inf').lower_bound
+        max_score = ScoreTest.decode(b'inf').upper_bound
+        items = list(zset.irange_score(min_score, max_score))
+        if count is None:
+            count = 1
+
+        items = self._limit_items(items, 0, count)
+        self.zrem(key, *[value for score, value in items])
+        items = self._apply_withscores(items, True)
+        return items
 
     def _zrange(self, key, start, stop, reverse, *args):
         zset = key.value
--- a/test/test_fakeredis.py
+++ b/test/test_fakeredis.py
@@ -2112,6 +2112,37 @@ def test_zadd_incr(r, ch):
     assert zadd(r, 'foo', {'three': 1.0}, incr=True, xx=True, ch=ch) == 4.0
 
 
+def test_zpopmin_empty(r):
+    assert r.zpopmin('foo') == []
+
+
+def test_zpopmin_default_count(r):
+    zadd(r, 'foo', {'one_a': 1})
+    zadd(r, 'foo', {'one_b': 1})
+    zadd(r, 'foo', {'two_a': 2})
+    assert r.zpopmin('foo') == [(b'one_a', 1.0)]
+    assert r.zcard('foo') == 2
+
+
+def test_zpopmin_zero_count(r):
+    zadd(r, 'foo', {'one_a': 1})
+    zadd(r, 'foo', {'one_b': 1})
+    zadd(r, 'foo', {'two_a': 2})
+    assert r.zpopmin('foo', count=0) == []
+    assert r.zcard('foo') == 3
+
+
+def test_zpopmin_more_than_available(r):
+    zadd(r, 'foo', {'one_a': 1})
+    zadd(r, 'foo', {'one_b': 1})
+    zadd(r, 'foo', {'two_a': 2})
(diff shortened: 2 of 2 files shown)


fakeredis/_server.py
```diff
@@ -2129,6 +2129,20 @@ def zincrby(self, key, increment, member):
     def zlexcount(self, key, min, max):
         return key.value.zlexcount(min.value, min.exclusive, max.value, max.exclusive)
 
+    @command((Key(ZSet),), (Int,))
+    def zpopmin(self, key, count=None):
+        zset = key.value
+        min_score = ScoreTest.decode(b'-inf').lower_bound
+        max_score = ScoreTest.decode(b'inf').upper_bound
+        items = list(zset.irange_score(min_score, max_score))
```
bmerry, line 2137: I don't think it's necessary to construct and pass `min_score` and `max_score`: you should be able to just pass `None` for both.

fakeredis/_server.py
```diff
@@ -2129,6 +2129,20 @@ def zincrby(self, key, increment, member):
     def zlexcount(self, key, min, max):
         return key.value.zlexcount(min.value, min.exclusive, max.value, max.exclusive)
 
+    @command((Key(ZSet),), (Int,))
+    def zpopmin(self, key, count=None):
```
bmerry, line 2133: Unfortunately the `@command` decorator doesn't support optional (0 or 1 instance) arguments yet, only repeating arguments (0 or more instances), so you need to make the argument a `*args` and then check in the function if there are too many arguments (see `_zrangebylex` for an example).

This will also need a unit test, which can be done with `r.execute_command` to directly pass parameters.

test/test_fakeredis.py
```diff
@@ -2112,6 +2112,37 @@ def test_zadd_incr(r, ch):
...
+def test_zpopmin_empty(r):
+    assert r.zpopmin('foo') == []
+
+
+def test_zpopmin_default_count(r):
+    zadd(r, 'foo', {'one_a': 1})
+    zadd(r, 'foo', {'one_b': 1})
+    zadd(r, 'foo', {'two_a': 2})
+    assert r.zpopmin('foo') == [(b'one_a', 1.0)]
+    assert r.zcard('foo') == 2
```
bmerry, line 2124: Rather test the actual contents (with `r.zrange(0, -1)`) to ensure that the right elements were removed.

test/test_fakeredis.py
```diff
@@ -2112,6 +2112,37 @@ def test_zadd_incr(r, ch):
...
 
 
+def test_zpopmin_empty(r):
+    assert r.zpopmin('foo') == []
+
+
+def test_zpopmin_default_count(r):
+    zadd(r, 'foo', {'one_a': 1})
+    zadd(r, 'foo', {'one_b': 1})
+    zadd(r, 'foo', {'two_a': 2})
```
bmerry, line 2122: Could do all three steps in one go, but more importantly, it would be better to use data where the elements with the lowest scores are not the ones that are first lexicographically, to ensure that it is really scores being considered.

test/test_fakeredis.py
```diff
@@ -2112,6 +2112,37 @@ def test_zadd_incr(r, ch):
...
+    assert r.zpopmin('foo') == [(b'one_a', 1.0)]
+    assert r.zcard('foo') == 2
+
+
+def test_zpopmin_zero_count(r):
+    zadd(r, 'foo', {'one_a': 1})
+    zadd(r, 'foo', {'one_b': 1})
+    zadd(r, 'foo', {'two_a': 2})
+    assert r.zpopmin('foo', count=0) == []
+    assert r.zcard('foo') == 3
```
bmerry, line 2132: Again, test the contents.

test/test_fakeredis.py
```diff
@@ -2112,6 +2112,37 @@ def test_zadd_incr(r, ch):
...
+    zadd(r, 'foo', {'two_a': 2})
+    assert r.zpopmin('foo') == [(b'one_a', 1.0)]
+    assert r.zcard('foo') == 2
+
+
+def test_zpopmin_zero_count(r):
+    zadd(r, 'foo', {'one_a': 1})
+    zadd(r, 'foo', {'one_b': 1})
+    zadd(r, 'foo', {'two_a': 2})
+    assert r.zpopmin('foo', count=0) == []
```
bmerry, line 2131: Can you also test with `-1` (`pytest.mark.parametrize` is good for this).

test/test_fakeredis.py
```diff
@@ -2112,6 +2112,37 @@ def test_zadd_incr(r, ch):
     assert zadd(r, 'foo', {'three': 1.0}, incr=True, xx=True, ch=ch) == 4.0
 
 
+def test_zpopmin_empty(r):
+    assert r.zpopmin('foo') == []
```
bmerry, line 2116: Can you also add a test that doesn't specify a count but for which the set is not empty, to test the code that defaults it to 1.

test/test_fakeredis.py
```diff
@@ -2112,6 +2112,37 @@ def test_zadd_incr(r, ch):
     assert zadd(r, 'foo', {'three': 1.0}, incr=True, xx=True, ch=ch) == 4.0
 
 
+def test_zpopmin_empty(r):
```
bmerry, line 2115: It looks like all the unit tests will need a `@redis3_only` decorator since redis-py 2.x didn't implement the command.

bmerry:
Fakeredis is under new maintenance at https://github.com/dsoftwareinc/fakeredis-py. I am thus closing all PRs on this repository. If you feel your PR is still relevant, please open a new one on that repository.

PR #316, opened 2021-11-07
Updates to support newer redis and redis-py

bmerry:
There are a few sorts of changes here:
- new commands (or options on commands) needed because redis-py 4.0 pre-releases depend on them
- other adjustments needed for compatibility with redis-py 4.0 pre-releases.
- tweaks to command behaviour to be compatible with redis 6.2.6 on some corner cases where it behaves differently to 6.0.10.

It doesn't work with redis-py 4.0.0rc1 due to https://github.com/redis/redis-py/issues/1669, but it should be in a much better position once redis-py gets rid of various assumptions of a 6.2 server.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,32 +9,32 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


ludwigschwardt (approved):
Looks good!

PR #318, opened 2021-11-15
Prepare a 1.7.0 release

bmerry:


bmerry:
@ludwigschwardt hold off on this if you haven't started yet - it will need an update after #319 is merged.

bmerry:
It seems like there is at least one more behavioural corner case that hypothesis is picking up. I'll have to get back to this next Monday.

bmerry:
@ludwigschwardt I think I've finally got all the tests to pass, so please take a look now.

fakeredis/_server.py
```diff
@@ -2164,14 +2164,17 @@ def zlexcount(self, key, min, max):
 
     def _zrange(self, key, start, stop, reverse, *args):
         zset = key.value
-        # TODO: does redis allow multiple WITHSCORES?
-        if len(args) > 1 or (args and not casematch(args[0], b'withscores')):
-            raise SimpleError(SYNTAX_ERROR_MSG)
+        withscores = False
+        for arg in args:
```
ludwigschwardt, line 2168: Do you test this with multiple args somewhere?

test/test_fakeredis.py
```diff
@@ -719,6 +725,16 @@ def test_set_ex_using_timedelta(r):
     assert r.get('foo') == b'bar'
 
 
+def test_set_ex_overflow(r):
+    with pytest.raises(ResponseError):
+        r.set('foo', 'bar', ex=18446744073709561)
```
ludwigschwardt, line 730: I would repeat the `# Overflows long long in ms` comment (in case the other test is deleted some day).

test/test_fakeredis.py
```diff
@@ -719,6 +725,16 @@ def test_set_ex_using_timedelta(r):
...
+
+
+def test_set_px_overflow(r):
+    with pytest.raises(ResponseError):
+        r.set('foo', 'bar', px=2**63 - 2)
```
ludwigschwardt, line 735: Does this also overflow? Why a different value?

test/test_fakeredis.py
```diff
@@ -709,6 +709,12 @@ def test_setex_using_float(r):
         r.setex('foo', 1.2, 'bar')
 
 
+@pytest.mark.min_server('6.2')
+def test_setex_overflow(r):
```
ludwigschwardt, line 713: What is the difference between SETEX and SET_EX?

ludwigschwardt (changes requested):
Minor suggestions - take it or leave it :-)

test/test_fakeredis.py
bmerry: setex corresponds to the redis SETEX command. set_ex corresponds to the redis SET command with the EX option.

fakeredis/_server.py
bmerry: The Hypothesis-based fuzz tester does, which is what prompted this. I haven't bothered writing a targeted test since it's not something I expect any code in the wild to actually do.

ludwigschwardt (approved):
👍🏾

bmerry:
I'm running into yet another test failure, but I'll raise a new PR for it and merge this so long.

PR #319, opened 2021-11-15
Add support for GET option to SET

bmerry:
I needed to implement this after switching to validating against a 6.2
Redis server, because the fuzzer in the hypothesis tests kept testing
that feature by accident.


ludwigschwardt (approved):
Looks reasonable.

PR #320, opened 2021-11-15
Fix an RPOP corner case that only worked for LPOP

bmerry:
Popping 0 elements from a list in Redis 6.2.6 is not an error even if
the key is the wrong type. This was correctly implemented for LPOP, but
for RPOP it wasn't done correctly (picked up by hypothesis).

### DIFF
--- a/fakeredis/_server.py
+++ b/fakeredis/_server.py
@@ -1853,7 +1853,7 @@ class FakeSocket:
                 key.update(new_value)
         return OK
 
-    @command((Key(list),), (Int(),))
+    @command((Key(),), (Int(),))
     def rpop(self, key, *args):
         return self._list_pop(lambda count: slice(None, -count - 1, -1), key, *args)
 


fakeredis/_server.py
```diff
@@ -1853,7 +1853,7 @@ def ltrim(self, key, start, stop):
                 key.update(new_value)
         return OK
 
-    @command((Key(list),), (Int(),))
```
ludwigschwardt, line 1856: Stupid question: What's the difference?

ludwigschwardt (approved):

fakeredis/_server.py
bmerry: With the old code, the wrapper will immediate raise an exception if the key exists and it is not a list before even trying to call the function. With the new code, the type is checked inside `_list_pop`, *after* some other checks.

ludwigschwardt: Thanks, thought as much!

PR #324, opened 2022-01-04
Add support for redis-py 4.1.0

rotten:
This pull request allows fakeredis to work with redis-py 4.1.0 which is the latest version at this time.

The only real difference in this new version is that the client sometimes sends a "disable_decoding" flag to the `read_response()` now which stops certain functions such as dump & restore from automatic decoding.  I added it as an optional parameter to the `FakeConnection.read_response` function.

All of the tests were passing for me after that change.


bmerry:
Thanks. I'm on holiday at the moment so will only be able to look at this next month.

bmerry:
Thanks for implementing this! I pushed a few small updates so that it will support 4.1.3, and will get it released now.

PR #328, opened 2022-03-19
Add Initial Support for Redis 4.2.0

jd014:
This pull request addresses breaking changes in redis 4.2.0. It does not add support for new features that redis 4.2.0 includes. 

The new [4.2.0rc1](https://github.com/redis/redis-py/releases/tag/v4.2.0rc1) release for redis integrates aioredis into the redis library for async support. As such, this PR drops aioredis as a dependency and uses the new async support that redis has built in. 

Changes:
- Replace all refs of aioredis with redis.asyncio (Keep naming convention of aioredis for simplicity)
- Update `read_response` in aioredis to match redis implementation
- Update argument inspection in `FakeRedisMixin` since redis 4.2.0rc1 broke it

TODO:
- Old redis versions (<=3.1) are failing tests in github actions and I cannot reproduce locally or figure out why they are failing. Is there still a need to support such old redis versions or can they be dropped?
- Update documentation to reflect these changes (I can help if needed). It seems like this should be a part of the next release preparation rather than this PR though.


bmerry:
Thanks, it looks like this was a fair amount of work. Unfortunately I'm no longer maintaining fakeredis - see the updated README.rst. Since I'm not the owner on PyPI I can't confer maintainership on anyone else, so your best bet may be to start a fork.

jd014:
That's a shame, but I understand. Thanks for the update. Will close out this PR in that case.

